PR #677, opened 2015-01-16
Fix build process for extra libraries

rolk:
The Python bindings loads libjob_queue (even if it is not used), which require BUILD_ERT to be set when building, which in turn builds all the libraries. These patches fixes some issues in those libraries, so they are at least buildable on Linux. (Tested on Ubuntu 12.04 Precise)


joakim-hove:
>The Python bindings loads libjob_queue (even if it is not used)

That was embarrassing - have made some attempts to make sure the pure ECLIPSE Python code can be used without the full spaceship, but we are clearly not there yet. Thanks for the heads up - will continue fiddling with this.

>These patches fixes some issues in those libraries

Improvements are always welcome.

PR #679, opened 2015-01-17
Use CMake command option to change directory

rolk:
Specifying the path without the -C option will give error messages like "git: 'devel/libenkf/applications/ert_tui' is not a git command.", but the -C option is not available in older Git versions (it was introduced in 1.8.5).

Use the WORKING_DIRECTORY option of the EXECUTE_PROCESS command instead to change into the source directory and run the git process there.


joakim-hove:
> Specifying the path without the -C option will give error messages like "git: 'devel/libenkf/applications/ert_tui' is not a git command."

Weird, without the `-C` this was in the venerable "Works for me" &#0153; state; maybe because I have the build directory under the git root? Anyway - this also works so I'm merging. Thank you.

PR #681, opened 2015-01-17
Take a stab at some of the warnings

rolk:
I believe the warnings generated by the original code here does not indicate any actual problem, so they should be removed to not drown out possibly important stuff.


joakim-hove:
The code base builds with 2-3 warnings on the "daily toolchain", which is gcc 4.1; but every time I try something newer I see that there are some warnings waiting out there. Thanks for the effort.

PR #731, opened 2015-03-20
ERT-861 Fix bug when TRANX not available, added default value

iLoop2:


joakim-hove:
Looks good - but should include a brief test just loading the offending dataset (mentioned in task description) - that task will currently bring the code to a complete failure.

joakim-hove:
> (mentioned in task description)

Now it is ...

joakim-hove:
In the current test the main export function - `ecl_nnc_export()` is never called for the problematic TYRIHANS case. I suggest adding one new test function:

```C

void test_nnc_export_missing_TRANX(const char * base ) {
    ecl_grid_type * grid = ...
    ecl_file_type * init_file = ...
    int export_size = ecl_nnc_export_get_size( grid );
    ecl_nnc_type * data = util_malloc( export_size * sizeof * data );

    test_assert_int_equal( ecl_nnc_export( grid , init_file , data ) , 0);
   free( data );
   ecl_file_close( init_file );
   ecl_grid_free( grid );
}

...

int main(int argc, char ** argv) {
    ...
    if (is_nnc_tran_null)
       test_nnc_export_missing_TRANX( base );
}
```

joakim-hove:
When a test actually calling `ecl_nnc_export()` succesfully with the problematic data set has been implemented, this can be merged.

PR #734, opened 2015-03-26
Ert 862 CustomKW support

jepebe:


joakim-hove:
This is superceeded by: #739

PR #750, opened 2015-04-21
ERT-866 TimeMap support in ErtServer

iLoop2:


joakim-hove:
Both the server and socket tests must be labeled with the StatoilData label.

joakim-hove:
Apart from minor comments this is ready to be merged!

PR #783, opened 2015-05-21
OPM-203 Added support for conversion from meters to feet for grid

iLoop2:


devel/libecl/include/ert/ecl/ecl_grid.h
```diff
@@ -172,7 +172,7 @@ extern "C" {
   void                    ecl_grid_grdecl_fprintf_kw( const ecl_grid_type * ecl_grid , const ecl_kw_type * ecl_kw , const char * special_header , FILE * stream , double double_default);
   bool                    ecl_grid_test_lgr_consistency( const ecl_grid_type * ecl_grid );
 
-  void                    ecl_grid_fwrite_EGRID(  ecl_grid_type * grid , const char * filename);
+  void                    ecl_grid_fwrite_EGRID(  ecl_grid_type * grid , const char * filename, bool is_metric);
```
joakim-hove: The variable should not be called `is_metric`; the grid itself *is* metric (or not), the input parameter here is the format you want to use for the output and a sensible name would be `metric_output` or `write_metric`.

devel/libecl/src/ecl_grid.c
```diff
@@ -5717,11 +5721,24 @@ static void ecl_grid_fwrite_EGRID__( ecl_grid_type * grid , fortio_type * fortio
   {
     {
       ecl_grid_assert_coord_kw( grid );
-      ecl_kw_fwrite( grid->coord_kw , fortio );
+      if(is_metric != grid->is_metric){
+          ecl_kw_type * coord_kw = ecl_kw_alloc_data_copy(grid->coord_kw);
```
joakim-hove: Here you should use the function `ecl_kw_alloc_copy()` which will create a full `ecl_kw_type` instance, the `ecl_kw_alloc_data_copy()` will only copy the internal *data* vector; not the header data.

devel/libecl/src/ecl_grid.c
```diff
@@ -5717,11 +5721,24 @@ static void ecl_grid_fwrite_EGRID__( ecl_grid_type * grid , fortio_type * fortio
...
+          ecl_kw_scale_float(coord_kw, METER_TO_FEET_SCALE_FACTOR);
```
joakim-hove: Since we are in this block we *know* that `output_metric != grid->is_metric` - i.e. a conversion should take place. However we must check one of the variables to know which way the conversion should go:

```
double scale_factor = if (grid->is_metric) ? METER_TO_FEET_SCALE_FACTOR : 1 / METER_TO_FEET_SCALE_FACTOR;
```

PR #786, opened 2015-05-26
fixed: probe for CXX11

akva2:
CXX wrapper uses std::shared_ptr with friends


joakim-hove:
Hmm - I guess your approach is superior; but have you looked at: https://github.com/Ensembles/ert/pull/781

akva2:
hrm, i guess i forgot to fetch origin. this can be closed.

PR #803, opened 2015-06-10
ERT-899: Fixed bug in socket error path

joakim-hove:


jepebe:
Pull Request Builder fails with this error: 

  File "/var/jenkins/workspace/ERT-PR-Test/build/python/tests/ert/server/test_socket.py", line 101, in test_connect
    self.runCommand( ["STATUS"] , port )
AssertionError: Exception not raised

PR #811, opened 2015-06-19
Ert 901 failed simulations

joakim-hove:
Report back failed simulations through the server.


devel/libjob_queue/src/job_queue_manager.c
```diff
@@ -78,10 +78,15 @@ int job_queue_manager_get_num_running( const job_queue_manager_type * manager) {
 }
 
 
-int job_queue_manager_get_num_complete( const job_queue_manager_type * manager) {
+int job_queue_manager_get_num_success( const job_queue_manager_type * manager) {
   return job_queue_get_num_complete( manager->job_queue );
```
flikka: Shouldn't the method job_queue_get_num_complete in job_queue also be renamed to num_success? To support the clarifying "theme" of this PR regarding going from "Complete" to "Success/Failed"?

flikka:
Apart from the comment on the naming of job_queue_get_num_complete - looks good.

The fact that these changes doesn't trigger changing of any tests, kind of indicate that we probably should have some more tests here. But the ones we have are kind of fragile, and we should perhaps re-think how we test these things (?)

agchitu:
Hi Kristian,
What is that status of this PR? I want to test this new version of the ert server.

flikka:
Hi Alin! I had some comments to it, but Joakim is on holiday now. Since it is important to you, and I only had minor comments, I can merge it later today.

- Kristian -

PR #817, opened 2015-07-02
A first draft for ert script interface to local config. 

arielalmendral:


joakim-hove:
Good - I will review this properly when holiday is over ~ 22.nd of July; however before that I would ask you to fixup this PR a bit:

1. In the `ERT-nnn` prefix the `nnn`points to an internal issue number, you can use `ERT-904`.
2. Ideally the work should be grouped in a couple of related commits, but at the very least the long list of commits with identical commit message should be squashed into one.

See `git interactive rebase`

arielalmendral:
Will do a cleanup and make new PR.

PR #818, opened 2015-07-14
Update URLs to the ERT wiki. 

perroe:
The wiki has moved, so the root address is now http://ert.nr.no/ert.


perroe:
Included in another pull-request.

PR #820, opened 2015-07-14
Handle client server communication issues

agchitu:
Any send-receive command that fails due to connection error or due to null response from the ERT server is resent after a 3 second sleep.


joakim-hove:
Thank you  - this looks very reasonable; I have installed a test version. 

The only thing I do not understand is that the situations now raising the `ConnectionErrorException` and `ServerReturnedNoneException` should previously have raised another exception, which should have bubbled all the way up to `seba_main` - and terminated the whole thing?

agchitu:
Hi Joakim. 
The request for simulations and in the end the exception happens on its own thread and it is not passed to the main calling thread. Only later when the results from the simulation are found missing seba_driver, dakota and seba_main will terminate.

joakim-hove:
Ok - I have read about threads and exceptions in Python. I have two comments: 

1. The final try: catch: should have a catch-all block, which should log that an unrecoverable exception has been caught.

2. Ideally the client should log more properly than just print(..), but I am happy to merge without.

agchitu:
Great, thanks. I'll write to Matthew and Lars that soon they can test the new version. There has been some confusion about which version to use. Initially I sent them the file by email (bad option I know) but I was not sure the solution is good enough and I wanted them to confirm. 

1. The solutions I found for capturing exceptions from children threads looked a bit more complicated but I did not test any yet. What I found interesting is that I could see that exceptions from more threads were logged.
2. All these print statements end up in the optimization report file so they were good enough for me. If you use a logger please print them as well  to either regular output or error streams so that the information still gets into the optimization files as well for the time being.

Will you close the PR or I need to do it? I forgot how this works.

joakim-hove:
We will not go into cross-thread catching, that is more complicated.

Print based logging is an ok start.

Bee cool if you update the final try Block, you can just push another commit to the branch in question (your master in this case) and I will merge it all.

agchitu:
Is have added the code, I hope this is what you meant.

PR #828, opened 2015-08-18
ERT-904: Local config work cont 1

FetKin:
Hi, Could you please integrate these two small changes?  Thanks, Ariel


joakim-hove:
Sure - they were quite innocent.

PR #829, opened 2015-08-24
Ert 906 Next Generation Plotting

jepebe:
Part of a longer job on converting the plotting functionality


devel/python/python/ert_gui/shell/shell_plot.py
```diff
@@ -16,6 +16,7 @@ class ShellPlot(object):
     def __init__(self, name):
         super(ShellPlot, self).__init__()
         clist = plt.rcParams['axes.color_cycle']
+        clist = ["#386CB0", "#7FC97F", "#FDC086", "#F0027F", "#BF5B17"]
```
joakim-hove: Can we have names here - possibly in a comment?

PR #831, opened 2015-08-25
ERT-904 Local script doc

FetKin:
Documentation of local config setup, both text based and python based.


arielalmendral:
Sorry for cross-user. User FetKin is wrong, the correct one should be arielalmendral.

PR #832, opened 2015-08-25
Ert 923 - Refcase and Observation plotting

jepebe:
Depends on ERT-906

### DIFF
--- a/devel/python/python/ert_gui/plottery/ensemble_plot.py
+++ b/devel/python/python/ert_gui/plottery/ensemble_plot.py
@@ -23,3 +23,21 @@ class EnsemblePlot(object):
         if len(lines) > 0:
             plot_config.addLegendItem(ensemble_label, lines[0])
 
+
+    @staticmethod
+    def plotRefcase(axes, plot_config, data):
+        """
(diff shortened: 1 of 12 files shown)


devel/python/python/ert_gui/plottery/ensemble_plot.py
```diff
@@ -0,0 +1,43 @@
...
+        line_marker = plot_config.lineMarker()
+        line_style = plot_config.lineStyle()
+        lines = axes.plot_date(x=data.index.values, y=data, color=line_color, alpha=line_alpha, marker=line_marker, linestyle=line_style)
+
+        if len(lines) > 0:
+            plot_config.addLegendItem(ensemble_label, lines[0])
+
+
+    @staticmethod
+    def plotRefcase(axes, plot_config, data):
```
joakim-hove: Since the `RefCase` really is about plotting summary vectors the `plotRefcase` method should maybe be in the `SummaryPlot` class instead?

joakim-hove:
With the exception of the refcase report_step / ministep issue this is ready to be merged.

PR #833, opened 2015-08-25
ERT-904 Local script doc

arielalmendral:
Documentation of local config setup, both text based and python based.


joakim-hove:
Looks good - merging.

arielalmendral:
Thanks!

PR #841, opened 2015-09-01
ERT-904: Local config work cont 2

arielalmendral:
Continuation local config work. Some new functions and documentation.

### DIFF
--- a/devel/python/python/ert/enkf/local_obsdata.py
+++ b/devel/python/python/ert/enkf/local_obsdata.py
@@ -52,6 +52,12 @@ class LocalObsdata(BaseCClass):
         already_exists_node_for_key = LocalObsdata.cNamespace().add_node(self, node)
         return already_exists_node_for_key
 
+    def delNode(self, key):
+        assert isinstance(key, str)
+        LocalObsdata.cNamespace().del_node(self, key)     
+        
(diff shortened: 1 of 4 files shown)


devel/python/python/ert/enkf/local_obsdata.py
```diff
@@ -52,6 +52,12 @@ def addNode(self, node):
         already_exists_node_for_key = LocalObsdata.cNamespace().add_node(self, node)
         return already_exists_node_for_key
 
+    def delNode(self, key):
```
jepebe: Suggestion:
Python has an operator **__delitem__(self, key)** similiar to **__getitem__**. 

Then you can delete items like this:
**del local_obsdata[key]**  

https://docs.python.org/2/reference/datamodel.html#emulating-container-types

jepebe: local_obsdata.py should also use registerObjectType

arielalmendral:
Thanks. Please let me know if your comments have been addressed.

jepebe:
Good. If you are going to continue working on the local_* classes I would really like to see more tests.

arielalmendral:
Thanks. There will be more tests coming.

PR #843, opened 2015-09-04
ERT-904: Local config tests

arielalmendral:
More local config tests as requested. Please note that this is ongoing work.


joakim-hove:
This looks good - will merge when the small items have been addressed.

arielalmendral:
Thanks and sorry for the delay, away from office. 

I think I have fixed this:

- Now reusing test context, if I understood correctly
- Possible to check number of Ministeps in UpdateStep
- Assert on retrieving Ministep.
- Avoid createTestPath.

joakim-hove:
Good - merging.

PR #846, opened 2015-09-07
Ert 916 obs consistency

joakim-hove:
Fix problem with obsevation <-> data consistency when reloading observations.

### DIFF
--- a/devel/python/python/ert/util/stringlist.py
+++ b/devel/python/python/ert/util/stringlist.py
@@ -69,6 +69,21 @@ class StringList(BaseCClass):
                     raise TypeError("Item: %s not a string" % s)
 
 
+    def __eq__(self , other):
+        if len(self) == len(other):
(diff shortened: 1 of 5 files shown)


devel/python/python/ert/util/stringlist.py
```diff
@@ -69,6 +69,21 @@ def __init__(self, initial=None):
                     raise TypeError("Item: %s not a string" % s)
 
 
+    def __eq__(self , other):
+        if len(self) == len(other):
+            if isinstance( other , StringList):
+                return StringList.cNamespace().equal(self, other)
+            else:
+                equal = True
+                for s1,s2 in zip(self , other):
```
jepebe: zip is not a generator so you end up with a complete list of tuples

joakim-hove:
Update pushed - using simple `enumerate` instead of `zip`

PR #853, opened 2015-09-16
fixed: install buffer_string.h

akva2:
installation broke a few weeks back with the introduction of buffer_string.h.


joakim-hove:
Thanks - merging.

PR #859, opened 2015-09-16
Linux warnings

perroe:
Fixed various warnings when compiling on Linux with gcc version 4.4.7 20120313 (Red Hat 4.4.7-16) (GCC).


joakim-hove:
Your approach of casting away 'const' is probably OK - but it looks a bit brutal? Is it kosher?

perroe:
As far as I know, casting away the const-ness i the standard way of doing this. The object itself is logically const, even though the lock objects have to be modified.

In C++ you also have the mutable keyword that is used for this kind of situations, but it is not available in C.

joakim-hove:
> In C++ you also have the mutable keyword that is used for this kind of situations, but it is not available in C.

Well - as I see it the C++ mutable keyword can be used to inform the compiler that the object is logically const although not bitwise const; in this case the object is still bitwise const; although something an element in the object points to is not bitwise const. But yes - I agree this should be OK - thanks for the tip.

PR #862, opened 2015-09-19
Added observations, ensemble and grid to local_config

joakim-hove:
__background:__ When creating a configuration for local updates observations and parameter which should be used are added based on string keys:

```Python
1:  local_config = ert.local_config()
2:  obs_set = local_config.createObsData("RandomObsKey")  
3:  obs_set.addNode("WWCT:OP_1")                        
4: 
5:  data_set = local_config.createDataSet("RandomDataKey")
6:  data_set.addNode("MULTFLT")
``` 

The strings used in line 3 and 6 above must correspond to valid observation and data keys respectively - however that is *not checked in any way*, and if invalid keys are used this will blow up hard in subsequent update step. The purpose of this PR is to check the keys used in the methodcalls in line 3 and 6 - and raise a `KeyError()` exception if the keys are invalid.

__Solution:__ The underlying C datastructure `local_config_type` does *not* have access to the ensemble configuration and observation datastructures, and giving that access becomes quite convoluted due to circular dependencies - it is therefor not possible to validate this at the lowest level. 

The solution chosen here is that the Python `LocalConfig` object is decorated with references to `EnsembleConfig` and `EnkfObs` - this works, but it breaks the pattern of "one Pythhon class <-> one C struct" - and it is not very clear to read from the `local_config.py` implementation that the `LocalConfig` instances *should* have members `ensemble_config, grid` and `obs`.

### DIFF
--- a/devel/python/python/ert/enkf/enkf_main.py
+++ b/devel/python/python/ert/enkf/enkf_main.py
@@ -102,7 +102,7 @@ class EnKFMain(BaseCClass):
         """ @rtype: Log """
         return EnKFMain.cNamespace().get_logh(self).setParent(self)
 
-    def local_config(self):
+    def getLocalConfig(self):
         """ @rtype: LocalConfig """
         config = EnKFMain.cNamespace().get_local_config(self).setParent(self)
         config.initAttributes( self.ensembleConfig() , self.getObservations() , self.eclConfig().get_grid() )
--- a/devel/python/python/ert/enkf/enkf_main.py
+++ b/devel/python/python/ert/enkf/enkf_main.py
@@ -104,8 +104,11 @@ class EnKFMain(BaseCClass):
 
     def local_config(self):
         """ @rtype: LocalConfig """
-        return EnKFMain.cNamespace().get_local_config(self).setParent(self)
-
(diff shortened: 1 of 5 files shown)


arielalmendral:
Your PR solves the data key verification issue at high level. However, the code becomes a bit unclear, especially if you offer functions like getObservations(). The user might get confused by which observations we talk about. 
 
I wonder if another possibility is allowing some optional “checkConsistency”, at the same level as “run”, that will run a check for internal consistency. The user needs to implement it. The check will consist in node checking, grid checking, etc… That way we just send data to local config, but local config does not own the data. Just a thought...

devel/python/python/ert/enkf/enkf_main.py
```diff
@@ -104,8 +104,11 @@ def logh(self):
 
     def local_config(self):
```
jepebe: This should be renamed localConfig or getLocalConfig

joakim-hove:
OK minor updates pushed:

1. Renamed `EnkFMain.local_config` -> `EnKfmain.getLocalConfig()`
2. Made the `getObservation()` and `getEnsembleConfig()` "private" with leading `__`

PR #864, opened 2015-09-23
ERT-904: Bug in ensemble update when using several ministeps

arielalmendral:
A local configuration was setup for testing of the scaling of observations.

The configuration consisted in one update step, and one ministep per cell of a porosity field. An eclipse region was defined to act as a mask: all cells of the mask were deactivated except for the cell that will be updated.
 
After running a Ensemble Smoother experiment, and comparing the updated field, we realized that only the last cell had been updated.

The investigation of this issue led us to the code in enkf_main_analysis_update, especifically, the loading and storing of parameters around the calls to enkf_main_serialize_dataset and enkf_main_deserialize_dataset.

In this case, the source file system (src_fs) is used to load data for each update, rewriting the previously updated cell in memory.

A workaround is to copy the data in the source file system to the target system.

We then use the same target filesystem to load and store for each local update. 

Please comment on this core fix. There are implications for iterative methods.

### DIFF
--- a/devel/libenkf/src/enkf_main.c
+++ b/devel/libenkf/src/enkf_main.c
@@ -1114,7 +1114,6 @@ static void enkf_main_analysis_update( enkf_main_type * enkf_main ,
                                        const local_ministep_type * ministep ,
                                        const meas_data_type * forecast ,
                                        obs_data_type * obs_data) {
-
   const int cpu_threads       = 4;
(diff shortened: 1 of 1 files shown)


arielalmendral:
Due to the implications for iterative methods, this fix needs to wait to be merged into master. Will reopen when conditions are created.

PR #865, opened 2015-09-24
Windows build

arielalmendral:


joakim-hove:
Thank you! I think we might drop this C89 requirement soon.

PR #871, opened 2015-10-01
Local updates

joakim-hove:
The three first commits make sure that a copy from src_case -> target_case for parameters which are not updated.

The last commits start the process of removing the updatestep completely.


devel/libenkf/src/local_config.c
```diff
@@ -531,13 +517,26 @@ struct local_config_struct {
...
+static local_updatestep_type * local_config_alloc_updatestep( local_config_type * local_config , const char * key ) {
+  local_updatestep_type * updatestep = local_updatestep_alloc( key );
+  hash_insert_hash_owned_ref( local_config->updatestep_storage , key , updatestep , local_updatestep_free__);
+  return updatestep;
+}
+
+
+
+void local_config_clear( local_config_type * local_config ) {
   local_config->default_updatestep  = NULL;
```
arielalmendral: This could potentially leak. Maybe safer to check that local_config->default_updatestep is not NULL and deallocate?

PR #874, opened 2015-10-12
ERT-904: The scaling of local observation std is truly now done

arielalmendral:


joakim-hove:
Closing this now - extended with test here: Ensembles#884

PR #882, opened 2015-10-19
fixed: respect DESTDIR

akva2:
DESTDIR is used by packaging tools

### DIFF
--- a/devel/cmake/python.cmake2
+++ b/devel/cmake/python.cmake2
@@ -8,7 +8,11 @@ function(add_python_package target package_path source_files install_package)
   foreach (file ${source_files} )
      set( source_file ${CMAKE_CURRENT_SOURCE_DIR}/${file} )
      set( build_file ${PROJECT_BINARY_DIR}/${package_path}/${file} )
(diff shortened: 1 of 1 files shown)


akva2:
bugger, scratch this one. tools only set destdir when doing make install so this won't help by itself.

akva2:
okay, got it. this suffices (in combination with packaging file updates).

joakim-hove:
> got it.

Well - I don't really get it. I thought `${CMAKE_INSTALL_PREFIX}` was the way to configure where your installation files should go - or is this an ability to configure a funny prefix path like `/tmp/package-build`? In that case I would assume a similar update must be applied to every single install target?

But sure - thanks I will merge when I am certain I have understood the situation.

akva2:
DESTDIR is an additional layer on top of the prefix. it's a 'prefix for the prefix'.
this is a common mechanism in both autotools and cmake. it's used to install *with* a given prefix, but into another folder. consider the packaging chroot - you don't want to install into the prefix path, but you want to package files with that prefix.

try it manually if you want. make install DESTDIR=/some/dir. you will get /some/dir/usr/local/<files> by default..

PR #884, opened 2015-10-19
Obs scale fixup

joakim-hove:
Fix bug in the scaling of observation standard deviation.


arielalmendral:
This looks very nice, thanks!

PR #886, opened 2015-10-20
update debian packaging

akva2:
- new package: python-ert.ecl
- set DESTDIR during configuration stage so correct paths are used to precompile python files.


joakim-hove:
Thank you!

PR #888, opened 2015-10-20
ERT-904: Check that local data/observation node already exists 

arielalmendral:


joakim-hove:
Will merge when the operator __contains__has been removed (or explained)

arielalmendral:
The line importing the  __contains__  operator was a mistake, now removed.

PR #889, opened 2015-10-20
Fix for eclipse 2015 file problem.

iLoop2:


joakim-hove:
OK good; this is close. Minor comments

* Not necessary to actually create a new test program for this - can just reuse the `ecl_grid_simple` test program:

```txt
add_test( ecl_grid_2015 ${EXECUTABLE_OUTPUT_PATH}/ecl_grid_simple ${PROJECT_SOURCE_DIR}/test-data/Statoil/....
```

* Add also a test loading the 2014.2 version of the file, so we do not regress on that.
* The tests must be marked with the `StatoilData` label

iLoop2:
Test is updated with label and reused the ecl_grid_simple test

PR #893, opened 2015-10-21
ERT-953 <CONFIG_FILE> and <CONFIG_FILE_BASE> are now magic strings in…

jepebe:
… config files


jepebe:
Debug info removed in ERT 950

joakim-hove:
> Debug info removed in ERT 950

OK - merging.

PR #895, opened 2015-10-21
Rename qc_module to hook_manager

arielalmendral:
First step towards introducing a "hook manager" that will administer the existing QC_WORKFLOW, and will be able to hook other HOOK_WORKFLOW.


arielalmendral:
OK, will be more precaucious in the future with  trailing white spaces. 
By the way, just learned that appending ?w=1 at the end of the URL gives a clean diff :)

PR #897, opened 2015-10-22
Ert 955 Plotting bugs and improvements

jepebe:


devel/python/python/ert/enkf/export/gen_data_observation_collector.py
```diff
@@ -33,44 +33,39 @@ def getObservationKeyForDataKey(ert, data_key, data_report_step):
         return observation_key
 
 
-
     @staticmethod
-    def loadGenDataObservations(ert, case_name, keys=None):
+    def loadGenDataObservations(ert, case_name, key=None):
```
joakim-hove: It seems `key = None` is not valid input; why is it optional at all?

devel/python/python/ert/enkf/export/gen_data_observation_collector.py
```diff
@@ -33,44 +33,39 @@ def getObservationKeyForDataKey(ert, data_key, data_report_step):
...
-
-            for iobs , (value,std) in enumerate(obs_node):
-                data_index = obs_node.getDataIndex( iobs )
-                data[key][data_index] = value
-                data["STD_%s" % key][data_index] = std
+        index_set = set()
+        obs_vector = enkf_obs[key]
+        report_step = obs_vector.activeStep()
+        obs_node = obs_vector.getNode(report_step)
+        """ :type: ert.enkf.observations.GenObservation """
```
joakim-hove: Hmmm - that was maybe pushing the IntelliJ features a bit far?

devel/python/python/ert/enkf/export/gen_data_observation_collector.py
```diff
@@ -33,44 +33,39 @@ def getObservationKeyForDataKey(ert, data_key, data_report_step):
...
+        for obs_index in range(len(obs_node)):
+            index_set.add(obs_node.getIndex(obs_index))
+
+        index_list = sorted(list(index_set))
+        data = DataFrame(index=index_list, columns=columns + std_columns)
+
+        for obs_index, (value, std) in enumerate(obs_node):
+            data_index = obs_node.getIndex(obs_index)
```
joakim-hove: I do not understand why this one is `data_index` - the purpose of the `data_index` is to find the right index in the data, i.e. `enkf_node` which we are observing - which can in general be different from the observation index. See: https://sdp.statoil.no/wikidocs/GBS/ERT/latest/user/observations/index.html#the-general-observation-keyword

devel/python/python/ert/enkf/key_manager.py
```diff
@@ -100,6 +101,23 @@ def genDataKeys(self):
...
+        """ :rtype: list of str """
+        if self.__gen_data_keys_with_observations is None:
+            enkf_obs = self.ert().getObservations()
+            gen_data_obs_keys = []
+            for obs_vector in enkf_obs:
+                report_step = obs_vector.activeStep()
+                key = obs_vector.getDataKey()
+
+                gen_data_key = "%s@%d" % (key, report_step)
+                if gen_data_key in self.genDataKeys():
```
joakim-hove: OK - I should have seen this before; but here is a key which does not actually exist in the `ensemble_config` proper - how wise is that?

devel/python/python/ert/enkf/export/gen_data_observation_collector.py
jepebe: Good point

devel/python/python/ert/enkf/export/gen_data_observation_collector.py
jepebe: That is not what I want. I want the observation value and its index in "GenData space" so that values are properly aligned when plotting along with normal GenData.

devel/python/python/ert/enkf/export/gen_data_observation_collector.py
jepebe: We can discuss this but this is just what the type hinting is for. For example: *getNode()* can return at least 3 different types of objects. The *:type:* expression just hints about it locally. Is this helpful for both reader and IDE? Yes. Is it pretty? Not really.

PR #908, opened 2015-10-28
Added test for ExtJob 

iLoop2:


devel/python/tests/ert/job_queue/test_ext_job.py
```diff
@@ -0,0 +1,27 @@
...
+                self.fail("ExtJob() raised Exception unexpectedly!")
+
+
+    def test_load_forward_model_with_error1(self):
+        with TestAreaContext("python/job_queue/workflow_job") as work_area:
+            self.assertRaises(ValueError, ExtJob,"dump_job",work_area.get_cwd(),False,"test")
+
+
+    def test_load_forward_model_with_error2(self):
+        with TestAreaContext("python/job_queue/workflow_job") as work_area:
```
jepebe: Prettier if these said forward_model instead of workflow_job

devel/python/tests/ert/job_queue/workflow_common.py
```diff
@@ -4,6 +4,11 @@
 class WorkflowCommon(object):
 
     @staticmethod
+    def createExternalDumpJobWithError():
+        with open("dump_job1", "w") as f:
+            f.write("EXECUTABLE dump1.py\n")
```
jepebe: what if you called this file dump_job_error?

PR #912, opened 2015-10-29
ERT-904: Checking that dataset and obsdata already exist in local config. Also obsset…

arielalmendral:
…->obsdata renaming


joakim-hove:
The exception `KeyError` should be used in the opposite situation - when a key does *not* exist; cool if you change `KeyError -> ValueError`. Then I will merge!

PR #914, opened 2015-10-29
Uppercase eclipse extensions

arielalmendral:
Allow the possibility of having lowercase extensions in Eclipse files.


joakim-hove:
OK - good; just add a comment that we ignore the "No miXed case rUlE" of Eclipse (which we did previously as well ... )

arielalmendral:
Not sure if I understood the comment correctly, tried to rephrase it.

PR #918, opened 2015-11-02
Default verbose info to true from Gert

arielalmendral:
The verbose information is very valuable to detect user errors and interpretation of results. Currently, the enkf->verbose is set to true in .../ert_tui/main.c line 161, for TUI, but not for GUI.

### DIFF
--- a/devel/python/python/ert/enkf/enkf_main.py
+++ b/devel/python/python/ert/enkf/enkf_main.py
@@ -32,8 +32,8 @@ def stringObj(c_ptr):
     return python_string
 
 class EnKFMain(BaseCClass):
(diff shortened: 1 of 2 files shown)


joakim-hove:
OK; I was a bit fast here. Could you use an environment variable e.g. `ERT_VERBOSE`:

```Python
verbose = False
verbose_var = os.getenv("ERT_VERBOSE")
if verbose_var.lower() == "true":
    verbose = True

```

arielalmendral:
Will do, in a new PR.

joakim-hove:
> Will do, in a new PR.

Thank you.

PR #920, opened 2015-11-03
Control verbose flag with environment variable

arielalmendral:
Here is the ERT_VERBOSE variable, related to #918.

Maybe the verbosity should be part of a logging system...


arielalmendral:
Default false done and tested.

joakim-hove:
Good - I am sorry to be so anal; now you are only missing the entry in 'release-notes.txt':
```txt
[Feature]  Set the the verbose flag from the environment variable ERT_VERBOSE.
```

arielalmendral:
Ahh, sorry, I thougth you were adding the release-notes. Sure that you want to have it under devel and not under docs?

joakim-hove:
> Sure that you want to have it under devel and not under docs?

Well I agree - `docs/` is better.

jepebe:
I am giving my vote for having it at the root. Easily accessible from the Ensmbles/ert front page.

joakim-hove:
> I am giving my vote for having it at the root. Easily accessible from the
> Ensmbles/ert front page.
>

OK - JeanPaul is stronger; at the *real root* - i.e. outside of devel/ as
well.

PR #921, opened 2015-11-03
Fix python wrappers on linux

akva2:
use versioned binaries in the python wrapper on relevant platforms
    
on linux, in particular, the libxxx.so is a symlink to the versioned object, and this symlink is only installed with the development packages. since we do not want to depend on the development packages to use the python bindings, this has been changed.

this PR sets the proper versions according to the ERT_VERSION_XXX variables, and adds the versioned binary name to the python wrappers.


joakim-hove:
OK - thanks; I guess this is a step in the right direction - but currently there are two problems with this:

1. The import machinery fails: https://travis-ci.org/Ensembles/ert/builds/88964623 - might be trivial have not had time to look properly on it.
2. When the `clib.py` file is made into a template it is impossible to run the Python distribution from source - which is unfortunate. Again - have not had time to look into alternatives; maybe a solution resembling the one used for the `__ert_lib_path.py`module could be used: here:https://github.com/Ensembles/ert/blob/master/devel/python/python/ert/__init__.py?

akva2:
1) some targets i haven't considered there i guess. shouldn't be hard to fix. i only tried summaryplot and that worked for me.

2) no, you cannot glob this thing, or well, you can glob but that voids the the whole point of versioned libraries. they are versioned so you have the ability to have different versions of a library on a system (see windows dll hell for the reason why). the information needs to be transferred from the build system somehow. if you want to keep the ability to run out of the source tree, one alternative is to stick the versions in some resource file, and read the resource file in clib.py, and then reimplementing the versions-to-soversion logic in the py file instead of the POSTFIX variable exported from the CMakeLists.txt

akva2:
1) can easily be cured by generating clib.py in the lib folder rather than the build folder (or as a consequence of doing the proposed changes in 2).

joakim-hove:
> 1) can easily be cured by generating clib.py in the lib folder rather than the build folder 

OK - I did not really understand; but you are probably right. As for running from the source tree I will not let that hold back merging now before the release. We can (hopefully) fix that again at a later stage.

akva2:
don't merge this is if you are not satisfied (which you shouldn't be). i'll just patch in the packaging until a proper solution has been found.

joakim-hove:
OK: slightly rewritten the second commit: https://github.com/Ensembles/ert/pull/923

joakim-hove:
Continued here: #923

PR #922, opened 2015-11-04
Final adjustments

akva2:
- depend on development package in python bindings
- bump versions

Please tag as release/2015.10/final when this has been merged.


joakim-hove:
OK - I am looking at : #921 now - I assume that also should go to the release branch before we can have a `2015.10/final` tag?

akva2:
no, let's not delay the release for that. the workaround is 6c8ef5c.

joakim-hove:
> let's not delay the release for that

Cool - merging; tag on the way.

joakim-hove:
Tag made.

PR #923, opened 2015-11-04
Fix python wrappers

joakim-hove:
This is an alternative version of #921


akva2:
looks good to me. too bad travis is having a bad day..

joakim-hove:
> looks good to me

I will close #921 - and then let this linger a short while; until we have and internal release out of the door.

> too bad travis is having a bad day..

Yes - those red 'X's look very bad!

PR #927, opened 2015-11-09
Checking validity of input/output transform

arielalmendral:
From the manual:

The options MIN, MAX, INIT_TRANSFORM and OUTPUT_TRANSFORM are all optional. MIN and MAX are as for dynamic fields. OUTPUT_TRANSFORM is the name of a mathematical function which will be applied to the field before it is exported, and INIT_TRANSFORM is the name of a function which will be applied to the fields when they are loaded. [Just use INIT_TRANSFORM:XXX to get a list of available functions.]

The option XXX seems not supported and it is uncertain which transformation ERT is actually using if the transform is not correct.

Suggesting to stop ERT if non-existing transformation is provided.

### DIFF
--- a/devel/libenkf/src/field_config.c
+++ b/devel/libenkf/src/field_config.c
@@ -455,6 +455,8 @@ static void field_config_set_init_transform( field_config_type * config , const 
   const char * init_transform_name = NULL;
   if (field_trans_table_has_key( config->trans_table , __init_transform_name))
     init_transform_name = __init_transform_name;
+  else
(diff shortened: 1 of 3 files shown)


joakim-hove:
Please: 

1. Add a simple function `field_trans_table_summary()` to print all the available functions before dying.
2. A line in the `release-notes.txt` file.

arielalmendral:
Safer to check if transformation is not NULL.

PR #935, opened 2015-11-12
ERT-961 Removed reporting with LaTeX

jepebe:


joakim-hove:
OK - will merge when it is back in the green

PR #944, opened 2015-11-20
Do not copy source fs into target fs if observations inactive

arielalmendral:
If one ministep had attached observations where all become unactive, all the previous ministep updates are ignored and the target is reset to the prior base case.

The source fs had been already copied to the target as the starting point, and we do the updates only on the target.


joakim-hove:
Can you give this: https://github.com/Ensembles/ert/pull/945 a spin - I would rather do it like that; if the results are the same.

arielalmendral:
The problem is the line calling "enkf_main_init_case_from_existing". That code is still triggered in #945

PR #951, opened 2015-12-03
Properly check return value of pthread_create()

joakim-hove:
Will properly check the return value of `pthread_create()` when adding test jobs.

### DIFF
--- a/devel/libjob_queue/tests/job_status_test.c
+++ b/devel/libjob_queue/tests/job_status_test.c
@@ -99,6 +99,11 @@ void test_update() {
         break;
     }
   }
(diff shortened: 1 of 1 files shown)


devel/libjob_queue/tests/job_status_test.c
```diff
@@ -66,28 +66,45 @@ void * user_done( void * arg ) {
...
+          num_done_threads++;
+        else
+          break;
+      }
+
+      i++;
+      if (i == N)
+        break;
+    }
   }
```
jepebe: Maybe check if num_done_threads + num_exit_threads > 0 ?

PR #952, opened 2015-12-03
Bug fixes and improvements

jepebe:
Generic stuff separated out from future ERT-977


joakim-hove:
Will merge when the micro issue in `ert_shell_context.py` has been fixed (or I am proved wrong ...)

PR #958, opened 2015-12-08
Hook manager continuation

arielalmendral:
First draft of hook manager with one hook.

1. The QC workflow lives a parallel life together with a generic hook workflow, for backward compatibility. The QC hook is now called (in the code) a post_hook_workflow. This hook has by definition the run mode = POST_SIMULATION. Previous functionality from QC workflow is supported, please let me know if some extended test fail. 

2. The hook workflow can be invoked from the NEW config keyword with either:

HOOK_WORKFLOW /path/to/hook_workflow.py PRE_SIMULATION

or

HOOK_WORKFLOW /path/to/hook_workflow.py POST_SIMULATION

3. A PRE_SIMULATION hook will be triggered immediately after starting a "Simulation". The POST_SIMULATION hook is triggered after the simulation, just as with the QC workflow.

4. Smoother, Iterative Smoother, Experiment, ES_MDA and Simulation support a hook.

5. Application: A local config workflow is typically a PRE_SIMULATION hook.


joakim-hove:
OK - I am looking into this; comments coming later today or tomorrow. Sorry it has taken so long.

devel/libenkf/src/hook_workflow.c
```diff
@@ -0,0 +1,224 @@
+/*
+   Copyright (C) 2012  Statoil ASA, Norway.
```
joakim-hove: 2015

arielalmendral:
Great, thanks!

devel/libenkf/src/hook_workflow.c
```diff
@@ -0,0 +1,224 @@
...
+#define QC_WORKFLOW_NAME    "QC WORKFLOW"
+#define RUNPATH_LIST_FILE   ".ert_runpath_list"
+
+
+struct hook_workflow_struct {
+  hook_run_mode_enum       run_mode;
+  char                   * path;
+  workflow_type          * workflow;
+  ert_workflow_list_type * workflow_list;
+  runpath_list_type      * runpath_list;
```
joakim-hove: There should only be one `runpath_list` instance - which should be managed by the `hook_manager`.

devel/libenkf/src/hook_workflow.c
```diff
@@ -0,0 +1,224 @@
...
+};
+
+
+hook_workflow_type * hook_workflow_alloc( ert_workflow_list_type * workflow_list , const char * path ) {
+  hook_workflow_type * hook_workflow = util_malloc( sizeof * hook_workflow );
+  hook_workflow->run_mode = PRE_SIMULATION;
```
joakim-hove: The manager should handle `run_mode` -  it should not be necessary to have that all the way down here?

devel/libenkf/src/hook_workflow.c
```diff
@@ -0,0 +1,224 @@
...
+#define RUN_MODE_POST_SIMULATION_NAME    "POST_SIMULATION"
+
+#define HOOK_WORKFLOW_NAME  "HOOK WORKFLOW"
+#define QC_WORKFLOW_NAME    "QC WORKFLOW"
+#define RUNPATH_LIST_FILE   ".ert_runpath_list"
+
+
+struct hook_workflow_struct {
+  hook_run_mode_enum       run_mode;
+  char                   * path;
```
joakim-hove: What is this `path` variable?

devel/libenkf/include/ert/enkf/hook_workflow.h
```diff
@@ -0,0 +1,62 @@
...
+extern "C" {
+#endif
+
+#include <ert/config/config_parser.h>
+
+#include <ert/enkf/ert_workflow_list.h>
+#include <ert/enkf/runpath_list.h>
+
+
+  typedef enum {PRE_SIMULATION  = 0,
```
joakim-hove: This enum defintion can (hopefully?) go to the manager implementation.

devel/libenkf/src/hook_manager.c
```diff
@@ -36,10 +36,8 @@
 
 
 struct hook_manager_struct {
-  char                   * path;
-  workflow_type          * workflow;
-  ert_workflow_list_type * workflow_list;
```
joakim-hove: The workflow_list is a shared resource and should be here?

devel/libenkf/src/hook_workflow.c
arielalmendral: Yes, the path is to support the old QC_PATH. I was under the impression that we wanted to maintain the old QC functionality as it was.

devel/libenkf/src/hook_workflow.c
arielalmendral: The idea here is that a manager is a collection of hooks and that each hook should know when it should be run.

devel/libenkf/include/ert/enkf/hook_workflow.h
arielalmendral: Yes, probably.

devel/libenkf/src/hook_manager.c
```diff
@@ -36,10 +36,8 @@
...
-  runpath_list_type      * runpath_list;
+  hook_workflow_type     * hook_workflow;
+  hook_workflow_type     * post_hook_workflow;
```
joakim-hove: I think these should be container types - probably `vector_type`. Maybe also a good idea to have a separate container manage the storage of the workflows, then the other lists can have the same reference multiple times.

joakim-hove:
All the internal tests passed!

joakim-hove:
Sorry this has taken so long - merging.

arielalmendral:
Thanks!

PR #967, opened 2015-12-20
C string

joakim-hove:
Small wrapper "class" to handle strings allocated in C and discarded in Python.

### DIFF
--- a/devel/python/python/ert/enkf/enkf_main.py
+++ b/devel/python/python/ert/enkf/enkf_main.py
@@ -22,14 +22,6 @@ from ert.enkf.key_manager import KeyManager
 from ert.util import SubstitutionList, Log
 
 
-# The method EnKFMain.fieldInitFile() allocates C storage for a char*;
-# the sole purpose of the stringObj method is to manage this memory.
-
-def stringObj(c_ptr):
-    char_ptr = ctypes.c_char_p( c_ptr )
-    python_string = char_ptr.value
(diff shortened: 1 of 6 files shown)


jepebe:
Since the string is a copy do you actually need the C-pointer? If you look in `enkf_main.py` a `string_obj` has already been implemented that converts the `char*` from C to a Python string and then frees the C-string.

joakim-hove:
> Since the string is a copy do you actually need the C-pointer?

No - it only used to free the C-storage, and then forgotten.

> If you look in enkf_main.py a string_obj has already been implemented that converts the char* from C to a Python string and then frees the C-string.

Ahhh - yes; I had a feeling we had quite recently implemented essentially this. I have changed the `EnkfMain` implementation to use the `CString` - which was essentially identical to the class previously found in `EnkFMain`.

jepebe:
In my opinion the solution in EnKFMain was cleaner since it was only a function and not a complete class. The result of the CString class is that you actually have no instances of that class. 

I think you can remove all of the class scaffolding in CString but keep your function that converts and handles Null pointer and then register that function as a type (as in EnkfMain solution). The result is that `cstring.py` only contains a function and type registering. 

Alternatively is is possible to integrate it into cwrap or as a part of `ert_util.__init__.py`.

joakim-hove:
OK - fixup pushed; will squash before merging.

jepebe:
Nice. :-)

PR #968, opened 2015-12-21
ERT-988: Removed all TUI plotting.

joakim-hove:
Removed the PLPLOT based plotting from TUI


jepebe:
FYI: I think there are some references to `plot config` (a class and some usage) and `plot path` in the Python tree. I think it is used in the GUI _somewhere_ (maybe saving of plots) and in the `plot settings` module in ertshell. If you need some proper file pointing, let med know.

joakim-hove:
> FYI: I think there are some references to plot config (a class and some usage) and plot path in the Python tree.

I could not find it - and now all tests are green (I think ...). Actually we had *two* `PlotConfig` classes!

jepebe:
`ert_gui/shell/plot_settings.py` line 39 refers to the ERT PlotConfig (PlotSettings get/setPath)

And the GUI usage for plot saving was off course removed when we changed the plotting engine.

joakim-hove:
> ert_gui/shell/plot_settings.py line 39 refers to the ERT PlotConfig (PlotSettings get/setPath)

OK - thanks got it; new attempt pushed.

PR #978, opened 2016-01-09
Reload obs2

joakim-hove:
Use the ert().loadObservations() method in ertshell - to ensure that local config is correctly updated.


devel/python/python/ert_gui/shell/observations.py
```diff
@@ -49,8 +49,8 @@ def load(self, config_file):
     @assertConfigLoaded
     def reload(self, config_file):
         if os.path.exists(config_file) and os.path.isfile(config_file):
-            self.ert().getObservations().clear()
-            self.ert().getObservations().load(config_file)
+            # If clear is False the new observations will be added to the existing.
+            self.ert().loadObservations( config_file , clear = True )
```
jepebe: Should also change the load shell function.

PR #986, opened 2016-01-14
Ert 982 geo

joakim-hove:


devel/python/tests/ert/server/test_simulation_context.py
```diff
@@ -55,3 +59,35 @@ def test_simulation_context(self):
                 else:
                     self.assertEqual(first_half_state_map[iens], RealizationStateEnum.STATE_INITIALIZED)
                     self.assertEqual(other_half_state_map[iens], RealizationStateEnum.STATE_HAS_DATA)
+
+
+    def test_runpath(self):
+        with ErtTestContext("ert/server/rpc/simulation_context_runpath", self.config2 , store_area = True) as test_context:
```
jepebe: store_area should probably be removed?

joakim-hove:
Fixup setting `store_area` to `False` is pushed.

PR #992, opened 2016-01-21
Posix spawn

joakim-hove:
This is work in progress. Have in particular *not* yet tested any other Python version than 2.7.6.


devel/libert_util/tests/ert_util_addr2line.c
```diff
@@ -91,7 +91,7 @@ int main( int argc , char ** argv) {
     
     util_chdir(path);
     dot_name = util_alloc_sprintf("./%s" , name);
-    child_pid = util_fork_exec( dot_name , 0 , NULL , true , NULL , NULL , NULL , NULL , NULL);
+    util_spawn(dot_name, 0, NULL, NULL, NULL, &child_pid);
```
jepebe: I do not think the child_pid signature is correct...

joakim-hove:
Fixed signature in `ert_util_addr2line()`

PR #999, opened 2016-01-28
Thin lifetime managed smspec_node

jokva:
These are simple C++-style automatically managed objects for handling
smspec_node with automatic deallocation and sugared constructors.

### DIFF
--- a/devel/CMakeLists.txt
+++ b/devel/CMakeLists.txt
@@ -127,7 +127,6 @@ if (MSVC)
    add_definitions( -D__func__="\\"????\\"")
 endif()
 
-
 if (ERT_LINUX)
    set( NEED_LIBM TRUE )
    set( LINK_STATIC FALSE )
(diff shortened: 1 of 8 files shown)


joakim-hove:
Will merge when the minor issues have been adressed / I have understood!

jokva:
It's not quite ready for merge yet, it's more of an early-submission-for-review.

jokva:
I'm fine with `smspec_node_set_num` being changed, and I don't even think it'll break any other client either, which is nice.

`struct deleter` is just a convenient public-by-default-class, so I won't need the `public` keyword for `deleter`. `deleter` is the mechanism `std::unique_ptr` uses to determine how to free its resource, which is a templated by-convention interface that `std::unique_ptr< T, deleter< T > >`'s T can be free'd by `deleter< T >::operator()( T* )`. For custom deleters, which we need in this case, some class need it's `operator()` implemented.

The library specifies the typedef, yes, but I struggled a bit with it's forward declaration. I can test some more - if it's convenient to remove the full use then I'll do it, but I seem to recall struggling a bit with the typedef as a template parameter.

PR #1000, opened 2016-01-29
Metacwrap conversion of ert.config

jepebe:


joakim-hove:
Very good - this wrapping is just getting better and better; will merge when the one funny character has been removed.

PR #1004, opened 2016-02-03
Using ert_unique_ptr for EclKW and FortIO.

joakim-hove:


devel/libeclxx/src/EclKW.cpp
```diff
@@ -4,20 +4,17 @@
 namespace ERT {
     template <>
     EclKW<int>::EclKW(const std::string& kw, int size_) {
-        ecl_kw_type * c_ptr = ecl_kw_alloc( kw.c_str() , size_ , ECL_INT_TYPE );
-        reset( c_ptr );
+        m_kw.reset( ecl_kw_alloc( kw.c_str() , size_ , ECL_INT_TYPE ) );
```
jokva: I suggest using the constructor directly here, instead of reset:

`: m_kw( ecl_kw_alloc( kw.c_str() , size_ , ECL_INT_TYPE )`

joakim-hove: I agree  - updated pushed.

PR #1006, opened 2016-02-03
Set default separator to ";" in the CSV export

Thif:


joakim-hove:
Very good - as you can see this PR actually fails a test. Go in here: https://github.com/Ensembles/ert/blob/master/devel/python/tests/core/ecl/test_sum.py#L56 and set the optional argument `delimiter = ";"` in the `csv.DictReader( )` construction.

joakim-hove:
Replaced by: https://github.com/Ensembles/ert/pull/1009

PR #1012, opened 2016-02-10
Ert 999 Cross Case Statistics Plot

jepebe:


joakim-hove:
Ready to merge when minor minor comments/questions have been adressed.

PR #1023, opened 2016-03-08
Reduce the use of current_case

joakim-hove:
It is a long term goal to completely remove the concept of *current case* from the enkf_main object. This PR is a step in that direction:

1. Have created a class `forward_load_context` which holds on to the information used while loading from the forward model. The most important aspect of this is that we avoid explicitly setting the current case variable of the `gen_data_config` instances. The `forward_load_context` implementation is currently very rough.

2. Have replaced many call to `enkf_main_get_fs()` deep down in the code with sending a `enkf_fs` parameter into the function in question - the case is seldom solved this way, when going outwards in the call stack we will typically still find a `enkf_main_get_fs()` call.

3. In the tui - and the workflow jobs have introduced functions `enkf_main_tui_get_fs()` and `enkf_main_job_get_fs()` - this changes nothing but hopefully makes the code slightly clearer.

The ambition for the current patch is to get the `forward_load_context` to work properly, and to move the `enkf_main_get_fs()`calls out as long as possible - the ambition is not to get completely rid of the current case - that will probably require a case manager in C.


devel/python/python/ert/enkf/forward_load_context.py
```diff
@@ -0,0 +1,50 @@
...
+# minimal; when creating the Python implementation the only purpose
+# was to get an existing test through.
+
+
+class ForwardLoadContext(BaseCClass):
+    TYPE_NAME = "forward_load_context"
+    _alloc = EnkfPrototype("void* forward_load_context_alloc( run_arg , bool , ecl_config , char* , stringlist )")
+    _select_step = EnkfPrototype("void forward_load_context_select_step( forward_load_context , int )")
+    _get_step = EnkfPrototype("int forward_load_context_get_load_step( forward_load_context)")
+    _free = EnkfPrototype("void forward_load_context_free( forward_load_context)")
```
jepebe: The C functions with forward_load_context as first argument should be bound.

devel/libenkf/include/ert/enkf/gen_data_config.h
```diff
@@ -63,7 +63,7 @@ typedef enum { GEN_DATA_UNDEFINED = 0,
   int                          gen_data_config_get_initial_size( const gen_data_config_type * config );
   void                         gen_data_config_assert_size(gen_data_config_type *  , int , int);
   const bool_vector_type     * gen_data_config_get_active_mask( const gen_data_config_type * config );
-  void                         gen_data_config_update_active(gen_data_config_type * config , int report_step , const bool_vector_type * data_mask);
+  void                         gen_data_config_update_active(gen_data_config_type * config , const forward_load_context_type * load_context , const bool_vector_type * data_mask);
```
jepebe: This is used by `test_gen_data_config.py` and` test_gen_data.py`

devel/python/tests/gui/ertshell/test_ertshell_smoother.py
```diff
@@ -6,45 +7,46 @@ class ErtShellSmootherTest(ExtendedTestCase):
 
     def test_smoother(self):
         test_config = self.createTestPath("local/custom_kw/mini_config")
-
-        with ErtShellTestContext("python/ertshell/smoother", test_config) as shell:
+    
+        with ErtShellTestContext("python/ertshell/smoother", test_config, store_area = True) as shell:
```
jepebe: remove `store_area`

devel/python/tests/gui/ertshell/test_ertshell_smoother.py
```diff
@@ -6,45 +7,46 @@ class ErtShellSmootherTest(ExtendedTestCase):
...
+    
             ert = shell.shellContext().ert()
             data = GenKwCollector.loadAllGenKwData(ert, "test_run", keys=["PERLIN_PARAM:SCALE"])
             update_data = GenKwCollector.loadAllGenKwData(ert, "test_run_update", keys=["PERLIN_PARAM:SCALE"])
-
+    
             self.assertTrue(data["PERLIN_PARAM:SCALE"].std() > update_data["PERLIN_PARAM:SCALE"].std())
 
 
-    def test_config(self):
```
jepebe: This test should not be disabled.

PR #1027, opened 2016-03-17
smspec_node supports copying

jokva:
Introduce a copy constructor for smspec_node for better C++ interop.
This needs a definition for a lot of C++ features to be well-formed,
even though it in most cases will be optimised away.

### DIFF
--- a/devel/libecl/include/ert/ecl/smspec_node.h
+++ b/devel/libecl/include/ert/ecl/smspec_node.h
@@ -103,6 +103,7 @@ typedef enum {ECL_SMSPEC_INVALID_VAR            =  0 ,
                                             float default_value);
 
   smspec_node_type *  smspec_node_alloc_new(int params_index, float default_value);
+  smspec_node_type *  smspec_node_alloc_copy( const smspec_node_type* );
 
   void                smspec_node_free( smspec_node_type * index );
   void                smspec_node_free__(void * arg);
--- a/devel/libecl/src/smspec_node.c
+++ b/devel/libecl/src/smspec_node.c
@@ -736,6 +736,40 @@ smspec_node_type * smspec_node_alloc_lgr( ecl_smspec_var_type var_type ,
   }
 }
 
+smspec_node_type* smspec_node_alloc_copy( const smspec_node_type* node ) {
+    if( !node ) return NULL;
+
+    smspec_node_type* copy = util_malloc( sizeof * copy );
+    UTIL_TYPE_ID_INIT( copy, SMSPEC_TYPE_ID );
+    copy->gen_key1 = util_alloc_string_copy( node->gen_key1 );
+    copy->gen_key2 = util_alloc_string_copy( node->gen_key2 );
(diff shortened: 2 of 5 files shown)


devel/libecl/src/smspec_node.c
```diff
@@ -736,6 +736,38 @@ smspec_node_type * smspec_node_alloc_lgr( ecl_smspec_var_type var_type ,
   }
 }
 
+smspec_node_type* smspec_node_copy( const smspec_node_type* node ) {
```
joakim-hove: The ert code has as a convention that: "every single function which allocates memory should have *alloc* as part of the name" - so cool if you rename:  `smspec_node_alloc_copy( )`.

devel/libecl/src/smspec_node.c
```diff
@@ -736,6 +736,38 @@ smspec_node_type * smspec_node_alloc_lgr( ecl_smspec_var_type var_type ,
...
+    if( !node ) return NULL;
+
+    smspec_node_type* copy = calloc( 1, sizeof( smspec_node_type ) );
```
joakim-hove: The convention - for good or bad - is:
```C
smspec_node_type * copy = util_malloc( sizeof * copy );
```

devel/libecl/src/smspec_node.c
```diff
@@ -736,6 +736,38 @@ smspec_node_type * smspec_node_alloc_lgr( ecl_smspec_var_type var_type ,
...
+    UTIL_TYPE_ID_INIT(copy, node->__type_id)
```
joakim-hove: Have tried to not access the `__type_id` variable explicitly; so would prefer initialization based on the #define symbol `SMPEC_TYPE_ID`

joakim-hove:
Good - will merge when we have agreed on the small details.

devel/libecl/src/smspec_node.c
jokva: I'll change the name, that's perfectly fine. I apologise for my lack of familiarity with ert here, obviously I should've studied it better.

devel/libecl/src/smspec_node.c
jokva: Was not aware.

devel/libecl/src/smspec_node.c
jokva: I'll look into reading the right ID then and copy it.

devel/libecl/include/ert/ecl/smspec_node.h
```diff
@@ -103,7 +103,9 @@ typedef enum {ECL_SMSPEC_INVALID_VAR            =  0 ,
                                             float default_value);
 
   smspec_node_type *  smspec_node_alloc_new(int params_index, float default_value);
+  smspec_node_type  * smspec_node_alloc_copy( const smspec_node_type* );
```
joakim-hove: I guess this header line should go - then we are ready to merge.

PR #1031, opened 2016-03-23
Create volume keyword from ecl_grid.

joakim-hove:
A utility to simplify / speed up volume summation.


jepebe:
1, 1A and tskille should maybe be revised in the future?

PR #1035, opened 2016-03-30
Fix build windows

arielalmendral:


arielalmendral:
- The current code does not compile with VS2012. Declaration fix in this PR.
- The code compiles with VS2015, possibly because it fully implements the C99 standard.

Do you still want to support VS2012?

devel/libecl/src/smspec_node.c
```diff
@@ -737,9 +737,11 @@ smspec_node_type * smspec_node_alloc_lgr( ecl_smspec_var_type var_type ,
 }
 
 smspec_node_type* smspec_node_alloc_copy( const smspec_node_type* node ) {
-    if( !node ) return NULL;
 
     smspec_node_type* copy = util_malloc( sizeof * copy );
+
+    if( !node ) return NULL;
```
joakim-hove: Thank you for the fix; I want to merge this - however the current fix will leak. I suggest instead:
```C
if (!node) return NULL;
{ 
     smspec_node_type * copy = util_malloc( sizeof * copy );
     ....
     ....
     return node;
}
}
```

arielalmendral:
OK, well observed. Using your fix...

joakim-hove:
Thank you - merging.

PR #1046, opened 2016-04-14
Ertshell Server Test Bug

jepebe:


devel/python/tests/gui/ertshell/ert_shell_test_context.py
```diff
@@ -33,7 +33,10 @@ def __init__(self, test_name, config_file, load_config=True, prefix=None, store_
     def __enter__(self):
         """ :rtype: ErtShell """
         test_area = self.test_area_context.__enter__()
-        test_area.copy_parent_content(self.config_file)
+
+        if os.path.exists(self.config_file):
```
joakim-hove: Shouldn't this rather raise `IOError`?

joakim-hove:
Should this fix the latest breaking in master?

jepebe:
Yes, I believe so.

PR #1047, opened 2016-04-14
Will raise IOError for invalid test_area input.

joakim-hove:


devel/python/python/ert/test/test_area.py
```diff
@@ -53,33 +53,67 @@ def get_original_cwd(self):
 
     def get_cwd(self):
         return self._get_cwd()
-        
+
+    def orgPath(self , path):
+        if os.path.isabs( path ):
+            return path
+        else:
+            return os.path.join( self.get_original_cwd( ) , path )
```
jepebe: I would also do `os.path.abspath()` before returning

joakim-hove: Well - the `self.get_original_cwd()` is already absolute.

jepebe: But your `path` variable is definitely relative at this point. It may be `ert/something` but it could also be `../ert/something`

joakim-hove: > but it could also be `../ert/something`

OK - update pushed.

PR #1048, opened 2016-04-14
don't use reserved names for the guard macros

andlaus:
it seems like the C specification reserves macros that begin and end
with two underscores for internal purposes: when compiling with
`clang++ -Weverything foo.cc` and if foo.cc includes e.g. the 'util.h'
ERT header, one gets

```
include/ert/util/util.h:20:9: warning: macro name is a reserved identifier [-Wreserved-id-macro]
```

(this was compiled using clang 3.6.)

since IMO the compiler is right about that complaint, let's just use a
different naming scheme for the guard macros of ert: this patch
replaces `__$FOO_H__` by `ERT_$FOO_H`.

### DIFF
--- a/devel/VisualStudio/stdbool.h
+++ b/devel/VisualStudio/stdbool.h
@@ -13,8 +13,8 @@
   include directory as used by VisualStudio. 
 */
 
-#ifndef __STDBOOL_H__
-#define __STDBOOL_H__
+#ifndef ERT_STDBOOL_H
+#define ERT_STDBOOL_H
 
 #ifndef __cplusplus
 typedef int bool;
--- a/devel/libanalysis/include/ert/analysis/analysis_module.h
+++ b/devel/libanalysis/include/ert/analysis/analysis_module.h
@@ -16,8 +16,8 @@
    for more details.
 */
 
-#ifndef __ANALYSIS_MODULE_H__
-#define __ANALYSIS_MODULE_H__
+#ifndef ERT_ANALYSIS_MODULE_H
+#define ERT_ANALYSIS_MODULE_H
 #ifdef  __cplusplus
 extern "C" {
 #endif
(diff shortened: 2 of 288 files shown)


joakim-hove:
I'm more than happy to squelch warnings - but it seems something has gone a little astray; travis is not entirely happy ....

andlaus:
> travis is not entirely happy .... 

I've seen it, but I'm a quite puzzled by this because there should not be anything different with this patch? (unless there is a naming conflict with an previously existing macro which I consider unlikely.)

devel/VisualStudio/stdbool.h
```diff
@@ -13,8 +13,8 @@
   include directory as used by VisualStudio. 
 */
 
-#ifndef __STDBOOL_H__
-#define __STDBOOL_H__
+#ifndef ERT_STDBOOL_H
+#define ERT_STDBOOL_H
```
andlaus: is it actually allowed to overwrite headers of the standard library? `stdbool.h` seems to be featured by the ISO-C standard...

andlaus:
weird: the error seems to be `ert/devel/libenkf/src/ert_template.c:52:36: error: unknown type name ‘ert_template_type’` but that type gets defined on my machine. is it possible that this is an issue with the ERT travis setup?

joakim-hove:
> is it possible that this is an issue with the ERT travis setup?

Everything is possible - will look at it tomorrow.

devel/VisualStudio/stdbool.h
joakim-hove: Probably not - at the time this written the Visual Studio C compiler was essentially K & R; I think things have improved since then.

joakim-hove:
Fixup here: https://github.com/Ensembles/ert/pull/1055

PR #1049, opened 2016-04-15
Windows ecl sum

joakim-hove:
Minor fixes for Windows.


joakim-hove:
Travis - get your act together :-(

joakim-hove:
OK - verified by windows users.

PR #1053, opened 2016-04-19
Adding PRE_SIMULATION hook support

arielalmendral:
Please have a look at the position of the presimulation hook and also some more phase steps that give more info. The presimulation hook is working with my test case. I could not find out why it was turned off in previous commits.

### DIFF
--- a/devel/docs/user/keywords/index.rst
+++ b/devel/docs/user/keywords/index.rst
@@ -1952,7 +1952,6 @@ run. Observe that the workflows being 'hooked in' with the
 :code:`HOOK_WORKFLOW` must be loaded with the :code:`LOAD_WORKFLOW`
 keyword.
 
-NB: Currently the :code:`PRE_SIMULATION` workflow is never called.
 
 Manipulating the Unix environment
 ---------------------------------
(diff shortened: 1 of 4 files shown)


joakim-hove:
Sorry for the slow review; I had to recall why I just removed the `PRE_SIMULATION` hook at the previous round. The order of things should be:

1. Create the runpath directories.
2. Run the `PRE_SIMULATION` hook.
3. Submit the simulations to the queue system.

With the current PR the `PRE_SIMULATION` hook will run *before* the runpath directories are created - and that is considerably less interesting. In the current code the steps 1 & 3 are run as one unit, so before we can get proper support for `PRE_SIMULATION` hooks we must split the running of simulations in a `init_eclipse` step and a 'submit jobs to queue' step. This is a nontrivial task of some complexity - I therefor suggest that is done first in a separate PR.

**Pointers:** 

1. This is the lowest level function which 
https://github.com/Ensembles/ert/blob/master/devel/libenkf/src/enkf_main.c#L1357 creates the runpath directories (function: `enkf_state_init_eclipse()`) 

2. This is the top level function: https://github.com/Ensembles/ert/blob/master/devel/libenkf/src/enkf_main.c#L1671
So - I *absolutely* want support for the `PRE_SIMULATION` hook, but then the first

joakim-hove:
Another comment: The current PR clearly shows that we should do some refactoring of the running of simulations from Python; the PR updates three essentially identical methods - this should be reorganized to one. But that can wait - the splitting of steps 1) and 3) above should get priority.

arielalmendral:
I see, thanks for the review! I will start this week with the suggested splitting.

PR #1055, opened 2016-04-21
Rename include guards

joakim-hove:
This is a minor fixup of: https://github.com/Ensembles/ert/pull/1048


andlaus:
thanks for looking into this and fixing the issue. I probably did not stumble over this error because I do not build the non-standard ERT applications and libraries on my machine. (because I only need ERT for OPM.)

joakim-hove:
> thanks for looking into this and fixing the issue.

Well - I'm the one who should say thanks! Merging now.

PR #1061, opened 2016-04-25
Fwd step enkf

arielalmendral:
Starting work on fwd_step_enkf.

1. Whitespace
2. A consistency check on number of folds with respect to number of ensembles.
3. A bug in function "regression_scale" in my opinion, since X_norm is set to be to the inverse of the standard deviation.

Let y = beta x + e be our original, unscaled regression problem. 

The scaled version y1 = beta1 x1 + e, where x1 = x/std(x) is not solved here since X is actually not being scaled even if the comment on the function says so. The solution beta1 to the scaled problem is std(x)*beta.

Instead, the original, unscaled regression is solved and the scaling is done on beta. To obtain the scaled version of beta, one multipies later on beta by the scaling std, in function "regression_unscale".

If I am correct, this bug is also affecting the method in lars.

After the proposed change on the scaling the updates now look reasonable.

### DIFF
--- a/devel/libanalysis/src/fwd_step_enkf.c
+++ b/devel/libanalysis/src/fwd_step_enkf.c
@@ -1,26 +1,26 @@
 /*
-   copyright (C) 2011  Statoil ASA, Norway. 
-    
-   The file 'fwd_step_enkf.c' is part of ERT - Ensemble based Reservoir Tool. 
-    
-   ERT is free software: you can redistribute it and/or modify 
-   it under the terms of the GNU General Public License as published by 
-   the Free Software Foundation, either version 3 of the License, or 
-   (at your option) any later version. 
-    
-   ERT is distributed in the hope that it will be useful, but WITHOUT ANY 
-   WARRANTY; without even the implied warranty of MERCHANTABILITY or 
-   FITNESS FOR A PARTICULAR PURPOSE.   
-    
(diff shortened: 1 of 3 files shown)


devel/libanalysis/src/fwd_step_enkf.c
```diff
@@ -97,6 +97,14 @@ void fwd_step_enkf_updateA(void * module_data ,
     int ens_size = matrix_get_columns( S );
     int nx = matrix_get_rows( A );
     int nd = matrix_get_rows( S );
+    int nfolds = fwd_step_data->nfolds;
+
+    /* Should this check be done in an internal module check? */
+    if ( ens_size <= nfolds){
+        fprintf(stderr, "**ERROR: The number of ensembles: %d must be larger than the CV fold %d\n", ens_size, nfolds);
+        fprintf(stderr, "**ERROR: Forward Stepwise regression failed...\n");
+        return;
```
joakim-hove: We can not really handle this failure; then I think it is better to just bomb out with `util_abort()`.

joakim-hove:
> After the proposed change on the scaling the updates now look reasonable.

I'll take your word for it - and merge when consistency check is enforced with `util_abort()`

arielalmendral:
Now aborting on error.

PR #1062, opened 2016-04-25
Added C++ implementation of test_work_area.

joakim-hove:


jokva:
Great addition, seems good.

PR #1064, opened 2016-04-26
Fwd step algorithm cleanup and improved output.

arielalmendral:
Only cosmetic changes.


joakim-hove:
OK; this looks good. Would be good if you added an internal setting to control the output. Add a bool option to the `fwd_step_enkf` implementation which can be controlled with:

```txt
ANALYSIS_SET_VAR FWD_STEP_ENKF VERBOSE TRUE
```

arielalmendral:
Supporting verbose.

joakim-hove:
Good - thank you; merging.

PR #1068, opened 2016-04-27
Adding PRE_SIMULATION hook support cont

arielalmendral:
I'd appreciate comments on this. This PR attempts to split as commented in:

#1053 

as some sort of replication of the way is done for tui when creating the run path. 

If this is OK, I'll add further support for PRE_SIMULATION to the tui functions.

### DIFF
--- a/devel/libenkf/src/enkf_main.c
+++ b/devel/libenkf/src/enkf_main.c
@@ -104,6 +104,7 @@
 #include <ert/enkf/analysis_iter_config.h>
 #include <ert/enkf/field.h>
 #include <ert/enkf/ert_log.h>
+#include <ert/enkf/ert_init_context.h>
 #include <ert/enkf/ert_run_context.h>
 #include <ert/enkf/run_arg.h>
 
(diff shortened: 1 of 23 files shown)


joakim-hove:
I'm looking on it.

devel/libenkf/src/enkf_main.c
```diff
@@ -1666,6 +1666,16 @@ void enkf_main_run_exp(enkf_main_type * enkf_main ,
...
 
 
+void enkf_main_run_create_run_path(enkf_main_type * enkf_main , bool_vector_type * iactive , int iter) {
+  init_mode_type init_mode = INIT_CONDITIONAL;
+  ert_run_context_type *run_context = enkf_main_alloc_ert_run_context_INIT_ONLY( enkf_main ,
+                                                                                enkf_main_get_fs( enkf_main ) ,
+                                                                                iactive ,
+                                                                                init_mode ,
+                                                                                iter );
+  enkf_main_run_exp(enkf_main , iactive , false );
```
arielalmendral: The idea is to put the code from this function  enkf_main_run_exp here.

arielalmendral:
This PR was for discussion only. We agree on doing a low level splitting, refactoring and eliminate INIT_ONLY as part of the work. Work coming in a different PR.

joakim-hove:
What is the status here? Is this WIP - or is it meant to be ready for review & merging?

arielalmendral:
WIP, fixing tests.

arielalmendral:
You may have a look at this attempt now. Please use the trick "?w=1" at the end of the URL to remove white space changes.

Note that I didn't touch the INIT_ONLY run context. It is bit too much for this iteration.

init_eclipse is now out of the queue system and create run path must be controlled by the user.

joakim-hove:
> Note that I didn't touch the INIT_ONLY run context. It is bit too much for this iteration.

No - I want that as well. This is a type of refactoring where the compiler helps you very much, just remove the enum definition and fix the compiler fallout.

devel/libenkf/src/enkf_main.c
```diff
@@ -1666,6 +1684,16 @@ void enkf_main_run_exp(enkf_main_type * enkf_main ,
 }
 
 
+void enkf_main_run_create_run_path(enkf_main_type * enkf_main , bool_vector_type * iactive , int iter) {
```
joakim-hove: That is one `run` too much: `enkf_main_create_run_path()` should suffice; if that creates conflicts downstream those functions should be renamed.

arielalmendral:
Oki, thanks.

arielalmendral:
As agreed, introduced a new "init_run_context" to handle the splitting, and cleanup of INIT_ONLY from run context. 

Following the same idea, one could introduce a "init_arg" to remove the INIT_ONLY dependency from run_arg?  I would prefer to do this in a separate PR if you think is OK.

And sorry for all these removed trailing whitespaces, they should be erradicated in ERT and forbidden.

joakim-hove:
> one could introduce a "init_arg" to remove the INIT_ONLY dependency from run_arg? I would prefer to do this in a separate PR if you think is OK.

Wait with that; the `run_arg` will have lifetime spanning *both* the initialisation and the simulation. For `run_context` we had two instances, one with initialisation lifetime, and one with simulation lifetime.

PR #1077, opened 2016-05-06
Adding innovations to forecast ensemble

arielalmendral:
This is another problem, apparent when testing fwd_enkf on parameters. The forecast ensemble was replaced by the innovations.

Need to merge for further testing. Still on testing phase, some of these WIP changes could be reverted.

### DIFF
--- a/devel/libert_util/src/regression.c
+++ b/devel/libert_util/src/regression.c
@@ -97,13 +97,27 @@ double regression_unscale(const matrix_type * beta , const matrix_type * X_norm 
    beta = inv(X'·X)·X'·y
 */
 
-void regression_OLS( const matrix_type * X , const matrix_type * Y , matrix_type * beta) {
+void regression_OLS( const matrix_type * X , const matrix_type * Y , const matrix_type* E, matrix_type * beta) {
+  int nvar = matrix_get_columns( E );
   matrix_type * Xt   = matrix_alloc_transpose( X );
-  matrix_type * Xinv = matrix_alloc( matrix_get_columns( X ) ,  matrix_get_columns( X ));
+  matrix_type * Xinv = matrix_alloc( nvar ,  nvar);
 
   matrix_matmul( Xinv , Xt , X );
-  matrix_inv( Xinv );
 
+  if (E){ // Add E'E to Xinv
+    matrix_type * Et  = matrix_alloc_transpose( E );
+    matrix_type * EtE = matrix_alloc( nvar ,  nvar);
+    matrix_matmul( EtE , Et , E );
+    matrix_inplace_add(Xinv, EtE);
+
+    for (int i = 0; i < nvar; ++i) // Sometimes the inversion fails, adding a small regularization
+        matrix_iadd(Xinv, i, i, 1e-10);
+
+    matrix_free( Et );
+    matrix_free( EtE );
+  }
+
+  matrix_inv( Xinv ); // Xinv is always invertible
   {
     matrix_type * tmp = matrix_alloc_matmul( Xinv , Xt );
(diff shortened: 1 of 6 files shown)


joakim-hove:
> Need to merge for further testing.

Why is that? We can install from an arbitrary version.

arielalmendral:
This can be reviewed and merged now.

joakim-hove:
> This can be reviewed and merged now.

Sorry - had not seen that; let us discuss it tomorrow.

arielalmendral:
Oki

devel/libert_util/src/regression.c
```diff
@@ -97,13 +97,27 @@ double regression_unscale(const matrix_type * beta , const matrix_type * X_norm
    beta = inv(X'�X)�X'�y
 */
 
-void regression_OLS( const matrix_type * X , const matrix_type * Y , matrix_type * beta) {
+void regression_OLS( const matrix_type * X , const matrix_type * Y , const matrix_type* E, matrix_type * beta) {
```
joakim-hove: I am quite skeptical to these changes which I feel are quite special case treatments added to the general regression function; if possible I would *really* prefer if the necessary fencing was done in the `fwd_step_enkf.c` implementation.

I do not really understand the linear algebra here, to me it seems like:
```latex
inv( X' X ) -> inv( X'X + E'E)? 
```

joakim-hove: If we ignore the origin in the `fwd_step_enkf` algorithm - what is the `E` matrix really in terms of this regression function?

arielalmendral:
Agree that this OLS is specific to fwd_step and it can be moved. The reason it was changed here was that the function was not in use elsewhere. The algebra is as you wrote: inv( X' X ) -> inv( X'X + E'E).

The idea is that this especial OLS will give the solution to the augmented regression problem:

[y' 0] = [Sc' E'] * beta + epsilon

devel/libert_util/src/stepwise.c
```diff
@@ -94,12 +99,8 @@ static double stepwise_estimate__( stepwise_type * stepwise , bool_vector_type *
     stepwise->X_norm = matrix_alloc( 1 , nvar );
 
     matrix_type * beta     = matrix_alloc( nvar , 1);           /* This is the beta vector as estimated from the OLS estimator. */
-    matrix_type * tmp_beta = matrix_alloc_copy( beta );         /* This is the beta vector shifted and scaled back to original variables. */
 
-    stepwise->Y_mean = regression_scale( X , Y , stepwise->X_mean , stepwise->X_norm );
-
-    regression_OLS( X , Y , beta );
-    y_mean = regression_unscale( beta , stepwise->X_norm , stepwise->X_mean , stepwise->Y_mean , tmp_beta );
+    regression_OLS( X , Y , E, beta );
```
joakim-hove: I need to understand exactly which regression equation is solved here?

joakim-hove:
> The reason it was changed here was that the function was not in use elsewhere.

yes - I see that; I though it was used here: https://github.com/Ensembles/ert/blob/master/devel/python/python/ert/util/stat.py - but that was wrong.

devel/libert_util/src/stepwise.c
joakim-hove: > [y' 0] = [Sc' E'] * beta + epsilon

What is the prime? Transpose?

pgdr: MATLAB / Octave uses prime for transpose, and this looks very MATLABy.

arielalmendral:
Renaming regression_OLS, use of generic Z input instead of E, and documentation.

PR #1081, opened 2016-05-10
Handle LOCAL_INACTIVE observations in output.

joakim-hove:


devel/libenkf/src/enkf_analysis.c
```diff
@@ -82,12 +82,17 @@ void enkf_analysis_fprintf_obs_summary(const obs_data_type * obs_data , const me
...
+              fprintf(stream , "           |");
             else if (active_mode == MISSING)
               fprintf(stream , "           |");
             else
               util_abort("%s: enum_value:%d not handled - internal error\n" , __func__ , active_mode);
+
             if (active_mode == MISSING)
               fprintf(stream , "                  Missing\n");
+            else if (active_mode == LOCAL_INACTIVE)
+              fprintf(stream , "                  Deactivated - local updates\n");
```
jepebe: What is deactivated here? The observation or the local updates?

joakim-hove: Using local updates parts of the observation has been deactivated.

PR #1084, opened 2016-05-11
Supporting plotting sigma 1, 2, and 3 for standard deviation

pgdr:
* Added function addSpinBox for CustomizationView
* Added std_dev_factor spinbox in StatisticsCustomizationView
* We now support sigma 1, 2, and 3 in CCS and Statistics


devel/python/python/ert_gui/tools/plot/customize/customization_view.py
```diff
@@ -69,7 +93,8 @@ def setter(self, style):
 
         self.updateProperty(attribute_name, getter, setter)
 
-
+    def updateStdDevFactor(self, value):
```
jepebe: I think this is unnecessary.

PR #1091, opened 2016-05-15
Created simple class perm_vector - with Python.

joakim-hove:
The purpose of the PermutationVector is to able to iterate over and
print the sort reordering in Python. This functionality is used in a workflow sorting realizations based on misfit.

Have also reorgainzed & updated the documentation of workflows - slightly.


devel/python/tests/core/util/test_vectors.py
```diff
@@ -18,13 +18,14 @@
 
 import copy
 import datetime
+import random
```
pgdr: Is this used?

joakim-hove:
> Is this used?¨

Ehhh - no. Updated pushed

devel/libert_util/include/ert/util/perm_vector.h
```diff
@@ -0,0 +1,32 @@
+/*
+   Copyright (C) 2016  Statoil ASA, Norway.
+
+   The file 'perm_vector.c' is part of ERT - Ensemble based Reservoir Tool.
```
pgdr: `.h`, if this line is necessary.

devel/libert_util/src/perm_vector.c
```diff
@@ -0,0 +1,76 @@
...
+  if (index < perm->size)
+    return perm->perm[index];
+  else {
+    util_abort("%s: invalid index:%d \n",__func__ , index );
+    return -1;
+  }
+}
+
+
+void perm_vector_iset( perm_vector_type * perm, int index, int value) {
```
pgdr: I see that this function might be nice to have, but isn't it somewhat "permissive"?  What I would expect would be that you have to take _two indices_ that you then flip.  But perhaps that becomes much harder to use.  I guess it's a trade-off here.

joakim-hove: > but isn't it somewhat "permissive"? What I would expect would be that you have to take two indices that you then flip

Yes - it is quite permissive; I tried to implement it as a swap with two indices - but those feeble attempts were quite far from O(1). This is the reason there is no `__setitem__` implemented in Python - I will at least add a comment.

devel/libert_util/src/perm_vector.c
```diff
@@ -0,0 +1,68 @@
...
+  perm_vector_type * perm = util_malloc( sizeof * perm );
+  UTIL_TYPE_ID_INIT( perm , PERM_VECTOR_TYPE_ID );
+  perm->size = size;
+  perm->perm = perm_input;
+  return perm;
+}
+
+
+void perm_vector_free( perm_vector_type * perm) {
+  free( perm->perm );
```
pgdr: Test if is `NULL`?

joakim-hove: > Test if is NULL?

`free( NULL )` is ok - I have learned.

PR #1093, opened 2016-05-18
Set indeterminate=False on running simulations

pgdr:
In 2951e31 `setPhaseName(name, indeterminate=True)` was added for the three simulators
* ensemble_experiment
* ensemble_smoother
* iterated_ensemble_smoother.

This led the progressbar to not display the progress correctly due to the simulations being indeterminate.

This patch sets `indeterminate=False`.


pgdr:
The function defaults `indeterminate` to `None`

    setPhaseName(name, indeterminate=None)

If preferred, we could remove the entire `indeterminate` from the function call.

PR #1095, opened 2016-05-20
Changed 'sigma' to 'Std dev multiplicative factor'

pgdr:
As per user request, `sigma` is now called `Std dev multiplicative factor`.


pgdr:
std-dev

pgdr:
std-dev

PR #1097, opened 2016-05-23
Added EnKFMain.smootherUpdate() method.

joakim-hove:
Small convenience method

### DIFF
--- a/devel/python/python/ert_gui/models/connectors/run/iterated_ensemble_smoother.py
+++ b/devel/python/python/ert_gui/models/connectors/run/iterated_ensemble_smoother.py
@@ -52,8 +52,9 @@ class IteratedEnsembleSmoother(BaseRunModel):
     def analyzeStep(self, target_fs):
         self.setPhaseName("Analyzing...", indeterminate=True)
         source_fs = self.ert().getEnkfFsManager().getCurrentFileSystem()
-        success = self.ert().getEnkfSimulationRunner().smootherUpdate(source_fs , target_fs)
+        es_update = self.ert().getESUpdate()
 
+        success = es_update.smootherUpdate(source_fs, target_fs)
         if not success:
             raise ErtRunError("Analysis of simulation failed!")
 
--- a/devel/python/python/ert_gui/models/connectors/run/multiple_data_assimilation.py
+++ b/devel/python/python/ert_gui/models/connectors/run/multiple_data_assimilation.py
@@ -15,6 +15,7 @@
 #  for more details.
 from ert.enkf.enums import EnkfInitModeEnum
 from ert.enkf.enums import HookRuntime
+from ert.enkf import ESUpdate
 
 from ert_gui.models.connectors.run import ActiveRealizationsModel,\
     TargetCaseFormatModel, AnalysisModuleModel, BaseRunModel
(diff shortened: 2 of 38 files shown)


jepebe:
In my opinion the EnkfSimulationRunner was the convenience location for running simulations and updates.

joakim-hove:
> In my opinion the EnkfSimulationRunner was the convenience location for running simulations and updates.

Well - then we disagree - at least as the only location. Running simulations and doing updates are *conceptually different things*. In particular the ability to perform updates without performing simulations is valuable.

jepebe:
This is how it is done for: MDA, ES, IES:
`success = self.ert().getEnkfSimulationRunner().smootherUpdate(source_fs , target_fs)`

joakim-hove:
> This is how it is done for: MDA, ES, IES:

Yes - I am not necessarily saying the existing codepath should be removed; but I insist that having the direct call on the `EnkFMain` class is valuable for script situations where you are not running simulations, only performing an update. For consistency I would be happy to remove the `smootherUpdate()` call from the simulation runner class.

jepebe:
My motivation for creating the `EnkfSimulationRunner` class was to group together this functionality together without bloating `EnkfMain`. There is no simulation in `EnkfSimulationRunner` when calling `smootherUpdate` except in the class name.

joakim-hove:
> [...] together without bloating EnkfMain.

Well - trying to avoid the bloat of `EnkFMain` is certainly a valid point.

> There is no simulation in EnkfSimulationRunner when calling smootherUpdate except in the class name.

As you point out yourself *running simulations* is a quite dominating part of the name though! Alternatively we could create a `ESUpdate` object.

jepebe:
Sounds good. Is it possible to have a class that can encompass both concepts? Like `EnkfSimulations`?

devel/python/python/ert_gui/models/connectors/run/iterated_ensemble_smoother.py
```diff
@@ -52,8 +52,9 @@ def createTargetCaseFileSystem(self, phase):
     def analyzeStep(self, target_fs):
         self.setPhaseName("Analyzing...", indeterminate=True)
         source_fs = self.ert().getEnkfFsManager().getCurrentFileSystem()
-        success = self.ert().getEnkfSimulationRunner().smootherUpdate(source_fs , target_fs)
-
+        
+        es_update = ESUpdate( self.ert() )
```
jepebe: Maybe use the same pattern as `getEnkFsManager()` and `getEnkfSimulationRunner()`

devel/python/python/ert_gui/models/connectors/run/multiple_data_assimilation.py
```diff
@@ -109,9 +110,11 @@ def update(self, target_case_format, iteration, weight):
 
         phase_string = "Analyzing iteration: %d with weight %f" % (next_iteration, weight)
         self.setPhase(self.currentPhase() + 1, phase_string, indeterminate=True)
-        self.ert().analysisConfig().setGlobalStdScaling(weight)
-        success = self.ert().getEnkfSimulationRunner().smootherUpdate(source_fs, target_fs)
 
+        es_update = ESUpdate( self.ert() )
+        self.ert().analysisConfig().setGlobalStdScaling(weight)
```
jepebe: Add `setGlobalStdScaling()` as a method on `ESUpdate`

PR #1100, opened 2016-05-26
EclKW::data()

jokva:


devel/libeclxx/include/ert/ecl/EclKW.hpp
```diff
@@ -77,6 +77,11 @@ namespace ERT {
             return *static_cast< T* >( ecl_kw_iget_ptr( this->m_kw, i ) );
         }
 
+        const typename std::remove_pointer< T >::type* data() const {
```
joakim-hove: This I do not understand; does this trickery by some magic give you *value* semantics with a copy? Or is this just a compilcated way of writing something like:
```C++
const T * data() const {
    return static_cast<const T*>( ecl_kw_get_ptr( this->m_kw ));
}
```

jokva: It's a trick to always give an array, even when the underlying data type is string `(char*)`.

The problem is that when we're storing constant-sized strings we're templating on `char*`, i.e. **with** the pointer type. If we'd template the return type here to be `T*` then `data` would return `char**`, which we can't do without either allocating or changing the underlying storage.

This makes the string-based return behave differently than its numeric siblings, but that's rather emphasised by the signature.

PR #1101, opened 2016-05-26
Ert 1046 crash

joakim-hove:
Make sure the job_queue is in state running before setting the user_exit flag

### DIFF
--- a/devel/libjob_queue/src/job_queue.c
+++ b/devel/libjob_queue/src/job_queue.c
@@ -855,7 +855,7 @@ void job_queue_run_jobs(job_queue_type * queue , int num_total_run, bool verbose
   int trylock = pthread_mutex_trylock( &queue->run_mutex );
   if (trylock != 0)
     util_abort("%s: another thread is already running the queue_manager\n",__func__);
-  else {
+  else if (!queue->user_exit) {
     /* OK - we have got an exclusive lock to the run_jobs code. */
 
     //Check if queue is open. Fails hard if not open
@@ -867,13 +867,13 @@ void job_queue_run_jobs(job_queue_type * queue , int num_total_run, bool verbose
       too many threads.
     */
     const int NUM_WORKER_THREADS = 4;
-    queue->running = true;
     queue->work_pool = thread_pool_alloc( NUM_WORKER_THREADS , true );
     {
       bool new_jobs         = false;
       bool cont             = true;
       int  phase = 0;
 
+      queue->running = true;
       do {
         bool local_user_exit = false;
         job_list_get_rdlock( queue->job_list );
@@ -1051,11 +1051,6 @@ void job_queue_start_manager_thread( job_queue_type * job_queue , pthread_t * qu
   arg_pack_append_int(queue_args  , job_size);
   arg_pack_append_bool(queue_args , verbose);
 
-  /*
-    The running status of the job is set to true here; this is to
-    guarantee that if calling scope queries the status of the queue
-    before queue_thread has actually started running the queue.
-  */
   job_queue->running = true;
   pthread_create( queue_thread , NULL , job_queue_run_jobs__ , queue_args);
 }
@@ -1259,12 +1254,33 @@ void job_queue_set_pause_off( job_queue_type * job_queue) {
 }
 
 /*
-  An external thread sets the user_exit flag to true, then subsequently the
-  thread managing the queue will see this, and close down the queue.
+  An external thread sets the user_exit flag to true, then
+  subsequently the thread managing the queue will see this, and close
+  down the queue. Will check that the queue is actually running before
+  setting the user_exit flag. If the queue does not change to running
+  state within a timeout limit the user_exit flag is not set, and the
+  function return false.
 */
 
-void job_queue_user_exit( job_queue_type * queue) {
-  queue->user_exit = true;
+bool job_queue_user_exit( job_queue_type * queue) {
+  if (!queue->user_exit) {
+    int timeout_limit = 10 * 1000000;  // 10 seconds
+    int usleep_time   =       100000;  // 0.1 second
+    int total_sleep   = 0;
+
+    while (true) {
+      if (queue->running) {
+        queue->user_exit = true;
+        break;
+    }
+      usleep( usleep_time );
+      total_sleep += usleep_time;
+
+      if (total_sleep > timeout_limit)
+        break;
+    }
+  }
+  return queue->user_exit;
 }
 
 
--- a/devel/libjob_queue/src/job_queue.c
+++ b/devel/libjob_queue/src/job_queue.c
@@ -1262,7 +1262,7 @@ void job_queue_set_pause_off( job_queue_type * job_queue) {
   function return false.
 */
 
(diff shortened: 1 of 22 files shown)


devel/python/python/ert/job_queue/queue.py
```diff
@@ -256,11 +256,19 @@ def set_max_running( self, max_running ):
...
-        if self.isRunning():
-            JobQueue.cNamespace().user_exit(self)
-
+        # The queue will not set the user_exit flag before the
+        # queue is in a running state. If the queue does not
+        # change to running state within a timeout the C function
+        # will return False, and that False value is just passed
+        # along.
+        user_exit = JobQueue.cNamespace().user_exit(self)
+        if user_exit:
```
jepebe: If user exit fails then the dialog will never be dismissed?

pgdr: I think I'm experiencing this right now.

devel/libjob_queue/src/job_queue.c
```diff
@@ -1249,15 +1244,38 @@ void job_queue_set_pause_off( job_queue_type * job_queue) {
...
+
+    while (true) {
+      if (queue->running) {
+        queue->user_exit = true;
+        break;
+    }
+      usleep( usleep_time );
+      total_sleep += usleep_time;
 
+      if (total_sleep > timeout_limit)
```
jepebe: Would it not be more natural to have:

`while(total_sleep <= timeout_limit)`

devel/python/python/ert/job_queue/driver.py
```diff
@@ -56,34 +80,36 @@ def submit( self, name, cmd, run_path, argList, num_cpu=1, blocking=False):
         argc = len(argList)
         argv = (ctypes.c_char_p * argc)()
         argv[:] = map(str, argList)
-        job_c_ptr = cfunc.submit(self, cmd, num_cpu, run_path, name, argc, argv)
+
+        job_c_ptr = self._submit(cmd, num_cpu, run_path, name, argc, argv)
+
         job = Job(self, job_c_ptr, blocking)
```
jepebe: I guess this method could return the job instance directly

devel/python/python/ert/job_queue/driver.py
```diff
@@ -56,34 +80,36 @@ def submit( self, name, cmd, run_path, argList, num_cpu=1, blocking=False):
         argc = len(argList)
         argv = (ctypes.c_char_p * argc)()
         argv[:] = map(str, argList)
-        job_c_ptr = cfunc.submit(self, cmd, num_cpu, run_path, name, argc, argv)
+
+        job_c_ptr = self._submit(cmd, num_cpu, run_path, name, argc, argv)
```
jepebe: Since you have reimplemented `job` then you can have the `submit` function return it directly

devel/python/python/ert/job_queue/__init__.py
```diff
@@ -74,8 +74,16 @@ def setenv( var, value):
...
+class QueuePrototype(Prototype):
+    lib = clib.ert_load("libjob_queue")
+
+    def __init__(self, prototype, bind=True):
+        super(QueuePrototype, self).__init__(QueuePrototype.lib, prototype, bind=bind)
+
+    
+#from .job_status_type_enum import JobStatusType
 from .job import Job
+from .job_status_type_enum import JobStatusType
```
jepebe: Switch order of these two

devel/python/python/ert_gui/models/connectors/run/multiple_data_assimilation.py
```diff
@@ -81,8 +81,21 @@ def runSimulations(self):
 
         for iteration, weight in enumerate(weights):
             self.simulateAndPostProcess(target_case_format, active_realization_mask, iteration)
+
+            # We exit because the user has pressed 'Kill all simulations'.
+            if self.userExitCalled( ):
+                print "User exit registered"
```
jepebe: Parenthesis

devel/libjob_queue/src/job_queue.c
joakim-hove: > Would it not be more natural to have:

In my opinion no - when there are multiple `break` statements in a loop I like to treat them equally; if the loop was implemented as:

```C
while (total_sleep <= timeout_limit ) { 
    if (queue->running) {
         break;
   }
   ....
}
``` 

The timeout limit and the `queue->running` tests would be treated differently.

joakim-hove:
OK  - fixups added for everything but the first comment.

devel/python/python/ert/job_queue/job.py
```diff
@@ -17,27 +17,30 @@
...
+
+    def free(self):
+        pass
 
 
     def block( self ):
         while True:
             status = self.status()
-            if status == STATUS_DONE or status == STATUS_EXIT:
+            if status == JobStatusType.STATUS_DONE or status == JobStatusType.STATUS_EXIT:
```
jepebe: Will fail without an import of `JobStatusType`. I am afraid that no test are directly testing the features in `Job`

devel/python/python/ert/job_queue/driver.py
```diff
@@ -16,21 +16,44 @@
...
+QueueDriverEnum.addEnum( "RSH_DRIVER" , 3 )
+QueueDriverEnum.addEnum( "TORQUE_DRIVER" , 4 )
+QueueDriverEnum.registerEnum(JOB_QUEUE_LIB, "queue_driver_enum")    
+
+
+class Driver(BaseCClass):
+    TYPE_NAME = "driver"
+    _alloc = QueuePrototype("void* queue_driver_alloc( queue_driver_enum )" , bind = False)
+    _set_option = QueuePrototype("void queue_driver_set_option( driver , char* , char*)")
+    _submit = QueuePrototype("Job queue_driver_submit_job( driver , char* , int , char* , char* , int , char**)")
```
jepebe: `Job` is actually called `job`

joakim-hove:
> OK - fixups added for everything but the first comment.

Again ....

devel/python/python/ert/job_queue/job.py
jepebe: And the state is called: `JOB_QUEUE_DONE` instead of `STATUS_DONE`

devel/python/python/ert/job_queue/driver.py
```diff
@@ -56,34 +80,34 @@ def submit( self, name, cmd, run_path, argList, num_cpu=1, blocking=False):
         argc = len(argList)
         argv = (ctypes.c_char_p * argc)()
         argv[:] = map(str, argList)
-        job_c_ptr = cfunc.submit(self, cmd, num_cpu, run_path, name, argc, argv)
-        job = Job(self, job_c_ptr, blocking)
+
+        job = self._submit(cmd, num_cpu, run_path, name, argc, argv)
```
jepebe: My mistake. `driver` and `queue_index` will be undefined if create like this.

joakim-hove: Ehhh - to tired; did not understand.

PR #1110, opened 2016-06-06
Windows build

arielalmendral:


joakim-hove:
Thank you;  but we are now targetting a newer version of VS and do not need this any longer. There is a automatic Windows build check attached to the pull requests.

arielalmendral:
OK, good to know :+1:

PR #1111, opened 2016-06-07
Analysis module logging

arielalmendral:
This addresses a current need for more detail output in analysis modules. In particular, fwd_step_enkf is a consumer.

The essential idea is the addition of an auxiliary struct that keeps track of the parameter blocks in the ensemble matrix for later reporting. The next step is to add support for data block reporting, possibly building on top of these structs.

Also in this PR you may find a log output functionality for fwd_step_enkf, very similar to rml_enkf.

I'd appreciate your opinion on these additions.

### DIFF
--- a/devel/libanalysis/include/ert/analysis/analysis_module.h
+++ b/devel/libanalysis/include/ert/analysis/analysis_module.h
@@ -96,7 +96,7 @@ typedef enum {
                                matrix_type * dObs ,
                                matrix_type * E ,
                                matrix_type * D ,
-                               module_info_type* module_info);
+                               const module_info_type* module_info);
 
 
   void                   analysis_module_init_update( analysis_module_type * module ,
--- a/devel/libanalysis/include/ert/analysis/analysis_module.h
+++ b/devel/libanalysis/include/ert/analysis/analysis_module.h
@@ -26,6 +26,7 @@ extern "C" {
 #include <ert/util/matrix.h>
 #include <ert/util/bool_vector.h>
 
+#include <ert/analysis/module_info.h>
 
 /*
    These are option flag values which are used by the core ert code to
@@ -94,7 +95,8 @@ typedef enum {
                                matrix_type * R ,
                                matrix_type * dObs ,
                                matrix_type * E ,
-                               matrix_type * D );
+                               matrix_type * D ,
+                               module_info_type* module_info);
 
 
   void                   analysis_module_init_update( analysis_module_type * module ,
--- a/devel/libanalysis/include/ert/analysis/analysis_table.h
+++ b/devel/libanalysis/include/ert/analysis/analysis_table.h
@@ -20,7 +20,7 @@ extern "C" {
                                          matrix_type * dObs ,
                                          matrix_type * E ,
                                          matrix_type * D ,
-                                         module_info_type* module_info);
+                                         const module_info_type* module_info);
 
 
   typedef void (analysis_initX_ftype)       (void * module_data ,
--- a/devel/libanalysis/include/ert/analysis/analysis_table.h
+++ b/devel/libanalysis/include/ert/analysis/analysis_table.h
@@ -10,25 +10,29 @@ extern "C" {
 #include <ert/util/rng.h>
 #include <ert/util/bool_vector.h>
 
-  typedef void (analysis_updateA_ftype) (void * module_data , 
-                                         matrix_type * A , 
-                                         matrix_type * S , 
-                                         matrix_type * R , 
-                                         matrix_type * dObs , 
+#include <ert/analysis/module_info.h>
+
+
+  typedef void (analysis_updateA_ftype) (void * module_data ,
+                                         matrix_type * A ,
+                                         matrix_type * S ,
+                                         matrix_type * R ,
+                                         matrix_type * dObs ,
                                          matrix_type * E ,
-                                         matrix_type * D );
-  
-  
-  typedef void (analysis_initX_ftype)       (void * module_data , 
-                                             matrix_type * X , 
-                                             matrix_type * A , 
-                                             matrix_type * S , 
-                                             matrix_type * R , 
-                                             matrix_type * dObs , 
-                                             matrix_type * E , 
+                                         matrix_type * D ,
+                                         module_info_type* module_info);
+
+
+  typedef void (analysis_initX_ftype)       (void * module_data ,
+                                             matrix_type * X ,
+                                             matrix_type * A ,
+                                             matrix_type * S ,
+                                             matrix_type * R ,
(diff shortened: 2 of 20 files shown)


devel/libanalysis/include/ert/analysis/analysis_module.h
```diff
@@ -112,6 +112,7 @@ typedef enum {
   const char           * analysis_module_get_table_name( const analysis_module_type * module);
   const char           * analysis_module_get_name( const analysis_module_type * module );
   bool                   analysis_module_check_option( const analysis_module_type * module , long flag);
+  void                   analysis_module_set_module_data_block_vector( analysis_module_type * module, void* arg );
```
joakim-hove: No - we should *not* set a module data object on the `fwd_step` structure - that should just be passed as an argument to the updateA routine.

joakim-hove: This `set` routine should be removed anyway - but why is the argument `void *arg`?

devel/libanalysis/include/ert/analysis/analysis_table.h
```diff
@@ -66,13 +67,14 @@ typedef struct {
   analysis_initX_ftype           * initX;
   analysis_init_update_ftype     * init_update;
   analysis_complete_update_ftype * complete_update;
+  analysis_module_data_block_vector_ftype * data_blocks;
```
joakim-hove: No - remove this; as long as possible the analysis module should be stateless - this state is owned an managed by calling scope.

devel/libanalysis/src/module_data_block.c
```diff
@@ -0,0 +1,84 @@
...
+
+void module_data_block_set_row_start(module_data_block_type * module_data_block, const int row_start ) {
+  module_data_block->A_row_start = row_start;
+}
+
+void module_data_block_set_row_end(module_data_block_type * module_data_block,  const int row_end ) {
+  module_data_block->A_row_end = row_end;
+}
+
+void module_data_block_set_active_indices(module_data_block_type * module_data_block, const int * index_list, const int n_active ) {
```
joakim-hove: Is it really necessary with all these `set_xxx` functions? I would assume it should be possible to get RAII behavior with something like:
```C
module_data_block_type * module_data_block_alloc( const char * key , const int * index_list , int row_start) {
    ...
}
```
and from then on the `module_block_data` object could be immutable.

devel/libanalysis/src/module_data_block_vector.c
```diff
@@ -0,0 +1,71 @@
...
+
+struct module_data_block_vector_struct {
+  UTIL_TYPE_ID_DECLARATION;
+  char                   * ministep_name;
+  vector_type            * data_block_vector;
+};
+
+
+module_data_block_vector_type * module_data_block_vector_alloc( const char* ministep_name) {
+  module_data_block_vector_type * module_data_block_vector = util_malloc( sizeof * module_data_block_vector );
```
joakim-hove: Missing: `UTIL_TYPE_ID_INIT( module_data_block_vector , MODULE_DATA_BLOCK_VECTOR_TYPE_ID );`

devel/libanalysis/src/module_data_block_vector.c
```diff
@@ -0,0 +1,71 @@
...
+}
+
+
+void module_data_block_vector_free( module_data_block_vector_type * module_data_block_vector ) {
+  util_safe_free(module_data_block_vector->ministep_name);
+  vector_free( module_data_block_vector->data_block_vector );
+  free( module_data_block_vector );
+}
+
+void module_data_block_vector_add_data_block( module_data_block_vector_type * module_data_block_vector , const char * datakey) {
```
joakim-hove: If possible the block structure should be completed in calling scope before it is added, instead of letting this routine create the block.

devel/libanalysis/src/fwd_step_log.c
```diff
@@ -0,0 +1,104 @@
...
+  FILE    * log_stream;
+};
+
+
+fwd_step_log_type * fwd_step_log_alloc() {
+  fwd_step_log_type * fwd_step_log = util_malloc( sizeof * fwd_step_log );
+  fwd_step_log->log_file = NULL;
+  fwd_step_log->log_stream = NULL;
+  fwd_step_log_set_clear_log( fwd_step_log , DEFAULT_CLEAR_LOG );
+  return fwd_step_log;
```
joakim-hove: Add a call: `fwd_step_log_set_log_file( fwd_step_log , DEFAULT_LOG_FILE);`

devel/libert_util/src/stepwise.c
```diff
@@ -457,11 +465,11 @@ void stepwise_free( stepwise_type * stepwise ) {
 
 }
 
-void stepwise_printf( const stepwise_type * stepwise , const int index ) {
+void stepwise_printf( const stepwise_type * stepwise , const char* key, const int active_index, const int global_index ) {
```
joakim-hove: Seems this has now aquired `fwd_step_enkf` specific arguments? I would prefer if:

1. The print functionality was moved to `fwd_step_log`.
2. The `stepwise_printf` function was just removed.

joakim-hove:
> The next step is to add support for data block reporting, possibly building on top of these structs.

If this step is in the pipeline already I would prefer if you made a parent struct (with no functionality) already in this PR - so that we would not to change datatypes more than necessary:
```
struct module_info_struct {
     module_data_block_vector_type * parameter_info;
     // To come: obs_data_info ?  
}
```

joakim-hove:
I am not entirely happy with the `module_data` name - but I currently do not have an alternative suggestion.

arielalmendral:
Thanks a lot for the comments. Changes coming soon...

devel/libanalysis/include/ert/analysis/analysis_module.h
```diff
@@ -94,7 +95,8 @@ typedef enum {
                                matrix_type * R ,
                                matrix_type * dObs ,
                                matrix_type * E ,
-                               matrix_type * D );
+                               matrix_type * D ,
+                               module_info_type* module_info);
```
joakim-hove: This can be `const module_info_type * module_info`.

joakim-hove:
OK - I will continue review tomorrow; but one immediate comment: The signature of calls in `AnalysisModule` must be updated. Is not necessary to implement the `module_info_type` in Python (altough that would be beneficial in the future) - just pass `None / NULL`.

arielalmendral:
Will do.

joakim-hove:
Better late than never - looking forward to get experience with this.

arielalmendral:
Thanks!

PR #1114, opened 2016-06-07
Added smspec_node query functions.

joakim-hove:


JacobStoren:
This looks like all that is needed from ResInsight.

PR #1116, opened 2016-06-14
Statoil testdata

joakim-hove:
Reorganized test data; should work nicely by default also when missing test data.


pgdr:
I'm unable to build (configure) with Statoil test data.

Specifically, ccmake dies when I try to add the path to the Statoil test data.

```cmake
 CMake Error at libenkf/tests/tests.cmake:1 (add_executable):
   add_executable cannot create target "enkf_runpath_list" because another
   target with the same name already exists.  The existing target is an
   executable created in source directory
   "/private/pgdr/ert/ert/devel/libecl/tests".  See documentation for policy
   CMP0002 for more details.
 Call Stack (most recent call first):
   libenkf/tests/CMakeLists.txt:1 (include)



 CMake Error at libenkf/tests/tests.cmake:2 (target_link_libraries):
   Attempt to add link library "enkf" to target "enkf_runpath_list" which is
   not built in this directory.
 Call Stack (most recent call first):
   libenkf/tests/CMakeLists.txt:1 (include)
```

joakim-hove:
OK - thank you. Fixup applied and force pushed. It work for me&trade;.

pgdr:
Thanks, that fixed it!

pgdr:
This is a nice change.  The testing works (and they all pass) on my machine both with the Statoil tests (464 tests) and without (294 tests, 293 on Travis). Travis is also very happy.

I cannot conscientiously say that I have reviewed this;  This is a 2000 lines diff of cmake scripts.  However, since this is purely cmake changes, I will merge it despite not having fully reviewed it.

devel/share/site-config
```diff
@@ -0,0 +1,7 @@
+-- This is a very basic site config file, mainly used for testing.
+
+WORKFLOW_JOB_DIRECTORY workflows/jobs/internal/config
+WORKFLOW_JOB_DIRECTORY workflows/jobs/internal-gui/config
+JOB_SCRIPT bin/job_dispatch.py
+QUEUE_OPTION LOCAL MAX_RUNNING 1
+
```
jepebe: Nice! :)

PR #1117, opened 2016-06-14
Removed EclQueue and ecl_local

pgdr:
Removed EclQueue

*  removed ecl_queue functionality and tests
* removed ecl_local


joakim-hove:
That was fast - some things missing:

1. Everything regarding the `ECL_LOCAL_TARGET` setting should be removed from [here](https://github.com/Ensembles/ert/blob/master/devel/python/python/ert/ecl/CMakeLists.txt#L42)
2. The Python class [`EclDefault`](https://github.com/Ensembles/ert/blob/master/devel/python/python/ert/ecl/ecl_default.py) is essentially a wrapper around the `ecl_local` module and can also be removed.

PR #1120, opened 2016-06-16
Obs data str

joakim-hove:


joakim-hove:
Should wait for research review for the second commit.

PR #1124, opened 2016-06-18
Create runpath file

joakim-hove:


devel/python/python/ert/enkf/runpath_list.py
```diff
@@ -1,36 +1,57 @@
 from collections import namedtuple
 from ert.cwrap import BaseCClass, CWrapper
```
jepebe: CWrapper should be removed

joakim-hove: > CWrapper should be removed

Fixed and force-pushed.

PR #1127, opened 2016-06-20
Removed MDA plugin

kjellkongsvik:
Is already a first class citizen


pgdr:
Wasn't this PR also supposed to change the labelling (GUI) for the MDA weights?

PR #1130, opened 2016-06-23
Vector operations on summary vectors

pgdr:
Added ability to scale and shift vectors

* Added `ecl_sum_scale_vector` and `EclSum.scaleVector`
* Added `ecl_sum_shift_vector` and `EclSum.shiftVector`
* removed 'const' from `ecl_sum_data_iget_ministep`
* added functions `ecl_sum_tstep_iscale` and `ecl_sum_tstep_ishift`
* added python tests for scaling and shifting vectors and key testing

### DIFF
--- a/devel/libecl/include/ert/ecl/ecl_sum.h
+++ b/devel/libecl/include/ert/ecl/ecl_sum.h
@@ -80,6 +80,7 @@ typedef struct ecl_sum_struct       ecl_sum_type;
   const char *     ecl_sum_iget_keyword( const ecl_sum_type * sum , int param_index );
   int              ecl_sum_get_data_length( const ecl_sum_type * ecl_sum );
   void             ecl_sum_scale_vector( ecl_sum_type * ecl_sum, int index, double scalar );
+  void             ecl_sum_shift_vector( ecl_sum_type * ecl_sum, int index, double addend );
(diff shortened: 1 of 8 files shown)


pgdr:
I did not add the `set()` feature, but that is easily addable now.  Should I do it?  How would that work?   Like `set(val) =?= shift( scale(key, 0), val)`

joakim-hove:
I agree that `set` is equal to `scale (0)` + `shift ()` -  leave it.

devel/libecl/include/ert/ecl/ecl_sum.h
```diff
@@ -79,6 +79,8 @@ typedef struct ecl_sum_struct       ecl_sum_type;
   const char *     ecl_sum_iget_wgname( const ecl_sum_type * sum , int param_index );
   const char *     ecl_sum_iget_keyword( const ecl_sum_type * sum , int param_index );
   int              ecl_sum_get_data_length( const ecl_sum_type * ecl_sum );
+  void             ecl_sum_scale_vector( ecl_sum_type * ecl_sum, int index, double scalar );
```
joakim-hove: `scale_factor` or `factor` would have been a better name than `scalar` - the fact that it scalar can actually just be read from the declaration, but says nothing of how it will be used. Anyway - just a minor minor detail; merging.

PR #1137, opened 2016-06-27
Make the failure to load SUMMARY <KEY> more verbose

flikka:
Changed the log message, and lowered the log level to highlight its importance.
Added some text to the error message in the python gui, with a reference to the ERT log file.

### DIFF
--- a/devel/python/python/ert/enkf/ert_log.py
+++ b/devel/python/python/ert/enkf/ert_log.py
@@ -1,24 +1,20 @@
-from ert.cwrap import CWrapper, CNamespace
-from ert.enkf import ENKF_LIB
+from ert.enkf import EnkfPrototype
 
 
 class ErtLog(object):
-    cnamespace = None
+    _init = EnkfPrototype("void ert_log_init_log(int, char*, bool)", bind=False)
+    _write_log = EnkfPrototype("void ert_log_add_message_py(int, char*)", bind=False)
+    _get_filename = EnkfPrototype("char* ert_log_get_filename()", bind=False)
 
-    @staticmethod
-    def log(log_level, message):
-        ErtLog.cnamespace.write_log(log_level, message)
+    @classmethod
+    def init(cls, log_level, log_filename, verbose):
+        cls._init(log_level, log_filename, verbose)
 
-    @staticmethod
-    def getFilename():
-        """ @rtype: int """
-        return ErtLog.cnamespace.get_filename()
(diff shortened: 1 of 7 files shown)


flikka:
Oh dear. Rusty. Fix coming up.

devel/python/python/ert/enkf/ert_log.py
```diff
@@ -6,12 +6,19 @@ class ErtLog(object):
...
+        return ErtLog.cnamespace.get_filename()
+
+
 ErtLog.cnamespace = CNamespace("ErtLog")
 
 cwrapper = CWrapper(ENKF_LIB)
 
 ErtLog.cnamespace.init = cwrapper.prototype("void ert_log_init_log(int, char*, char*, bool)")
 ErtLog.cnamespace.write_log = cwrapper.prototype("void ert_log_add_message_py(int, char*)")
+ErtLog.cnamespace.get_filename = cwrapper.prototype("char* ert_log_get_filename()")
```
jepebe: Upgrading the class to the new wrapping style would be a nice contribution.

jepebe: `runpath_list.py` is a good example of the changes necessary.

jepebe:
Will an unobserved Summary key still fail the simulation?

flikka:
Will an unobserved Summary key still fail the simulation; Yes, only changes are related to logging / error messages. Could try to change this, for example by modifying enkf_state_complete_forward_modelOK to allow this, since I believe there is a special flag set when SUMMARY keys fail to load. 

Can try to test that out, to see if it seems to be without (obvious) side-effects...

devel/python/python/ert/enkf/ert_log.py
flikka: Thanks, I'll do that!

joakim-hove:
> Could try to change this

That would be good. I think the changes should be [here](https://github.com/Ensembles/ert/blob/master/devel/libenkf/src/enkf_state.c#L593) - where one could test first whether the key in question was observed, before calling the `forward_load_context_update_result()` method:

```C
if (ensemble_config_has_node( ensemble_config , key)) {
   const enkf_config_node * config_node = ensemble_config_get_node( ensemble_config , key );
   if ( enkf_config_node_get_num_obs( config_node ) > 0) 
       forward_load_context_update_result(, LOAD_FAILURE);
   }
}
```

PR #1139, opened 2016-06-27
Set QT as backend to matplotlib

pgdr:
Set QT as backend to matplotlib

When importing matplotlib for the first time, set backend to
QT4.  If plot is imported or called prior to `matplotlib.use()`,
the `.use()` call will be _useless_.

### DIFF
--- a/devel/python/python/ert_gui/__init__.py
+++ b/devel/python/python/ert_gui/__init__.py
@@ -6,4 +6,14 @@ REQUIRED_VERSION_HEX = 0x02070000
 
 if sys.hexversion < REQUIRED_VERSION_HEX:
     raise Exception("ERT GUI Python requires at least version 2.7 of Python")
-    
+
+import os
+import matplotlib
+
+def headless():
+    return "DISPLAY" not in os.environ
+
+if headless():
+    matplotlib.use("Agg")
+else:
+    matplotlib.use("Qt4Agg")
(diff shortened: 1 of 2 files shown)


jepebe:
Add this to `ert_gui.__init__.py`:

```
import os
import matplotlib

def headless():
    return "DISPLAY" not in os.environ

if headless():
    matplotlib.use("Agg")
else:
    matplotlib.use("Qt4Agg")
```

And remove the same code from `ert_gui.plottery.plots.__init__.py` for a solution that covers **gert_main**, **ertshell** and **ertplot** and also supports running **headless**.

devel/python/python/ert_gui/shell/ertshell.py
```diff
@@ -2,6 +2,8 @@
 from cmd import Cmd
 import readline
 import os
+import matplotlib
+matplotlib.use('QT4Agg')
```
jepebe: This should not be necessary anymore.

devel/python/python/ert_gui/tools/plot/plot_widget.py
```diff
@@ -1,5 +1,7 @@
 import sys
 import traceback
+import matplotlib
+matplotlib.use('QT4Agg')
```
jepebe: This should not be necessary anymore.

devel/python/python/ert_gui/__init__.py
```diff
@@ -6,4 +6,14 @@
...
+import os
+import matplotlib
+
+def headless():
+    return "DISPLAY" not in os.environ
+
+if headless():
+    matplotlib.use("Agg")
+else:
+    matplotlib.use("Qt4Agg")
```
jepebe: You must also remove the same code from `ert_gui.plottery.plots.__init__.py`

PR #1142, opened 2016-06-28
ERT-1057 Improved MDA simulation panel

jepebe:
Added a "visualization" of the normalized weights.
Moved the help information into the help system.
Increased the robustness of weight parsing and calculation.


devel/python/python/ert_gui/models/connectors/run/multiple_data_assimilation.py
```diff
@@ -170,14 +171,16 @@ def normalizeWeights(weights):
     def parseWeights(weights):
         if not weights:
             return []
+
         elements = weights.split(",")
+        elements = [element.strip() for element in elements if not element.strip() == ""]
```
joakim-hove: Is the `if not element.strip() == ""` part required?

joakim-hove:
Looks good - happy to merge; possibly except the little question?!

PR #1145, opened 2016-06-28
Set indeterminate True.

joakim-hove:


devel/python/python/ert_gui/models/connectors/run/ensemble_smoother.py
```diff
@@ -44,7 +44,7 @@ def runSimulations(self):
         self.setPhaseName("Post processing...", indeterminate=True)
         self.ert().getEnkfSimulationRunner().runWorkflows( HookRuntime.POST_SIMULATION )
 
-        self.setPhaseName("Analyzing...", indeterminate=True)
+        self.setPhaseName("Analyzing...")
```
jepebe: I believe that setting Indeterminate is correct here.

devel/python/python/ert_gui/models/connectors/run/ensemble_smoother.py
```diff
@@ -55,10 +55,10 @@ def runSimulations(self):
         if not success:
             raise ErtRunError("Analysis of simulation failed!")
 
-        self.setPhase(1, "Running simulations...", indeterminate=False)
+        self.setPhase(1, "Running simulations...")
         self.ert().getEnkfFsManager().switchFileSystem(target_fs)
         
-        self.setPhaseName("Pre processing...", indeterminate=True)
```
jepebe: This one too.

PR #1146, opened 2016-06-28
Write __version__ number to ert_lib_info.

pgdr:
After compile,
* `python/ert/__ert_lib_info.py:3:__version__ = "2.0.git"`
* `python/ert_lib_info_install.py:3:__version__ = "2.0.git"`
will be written to `ert/build`.  The major.minor.micro is fetched from
cmake files `ert_lib_info_build.py.in` and corresponding install file.


pgdr:
As per [PEP-396](https://www.python.org/dev/peps/pep-0396/).

pgdr:
Correctly writes `__version__` to `ert` module.

joakim-hove:
Suggested test [here](https://github.com/Ensembles/ert/blob/master/devel/python/tests/core/util/test_version.py): 
```Python
def test_root_version(self):
   version = Version.currentVersion( )
   self.assertEqual( ert.__version__ , "%s.%s.%s" % (version.stringVersion() )
```

pgdr:
Thank you, Joakim.  I have added the test now and it even passes.

PR #1147, opened 2016-06-28
Added get->key1() to smspec_node C++

joakim-hove:


joakim-hove:
Self merging to get into OPM development.

PR #1150, opened 2016-06-30
Added python module check in cmake

pgdr:
* Checks if certain modules are present
* Checks if the version numbers are high enough
* If no PyQt4, then do not build ERT_GUI

### DIFF
--- a/devel/CMakeLists.txt
+++ b/devel/CMakeLists.txt
@@ -17,17 +17,44 @@ set( ERT_VERSION_MICRO git )
 #-----------------------------------------------------------------
 
 
-option( BUILD_ERT           	"Build the full ERT application - Linux only" OFF)
-option( BUILD_TESTS         	"Should the tests be built" OFF)
-option( BUILD_APPLICATIONS  	"Should we build small utility applications" OFF)
-option( BUILD_ECL_SUMMARY   	"Build the commandline application ecl_summary" OFF)
-option( BUILD_PYTHON        	"Run py_compile on the python wrappers" ON)
-option( BUILD_SHARED_LIBS   	"Build shared libraries" ON )
-option( INSTALL_ERT         	"Should anything be installed when issuing make install?" ON)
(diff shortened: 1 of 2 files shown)


devel/CMakeLists.txt
```diff
@@ -17,17 +17,73 @@ set( ERT_VERSION_MICRO git )
...
-option( BUILD_PYTHON        	"Run py_compile on the python wrappers" ON)
-option( BUILD_SHARED_LIBS   	"Build shared libraries" ON )
-option( INSTALL_ERT         	"Should anything be installed when issuing make install?" ON)
-option( ERT_BUILD_GUI       	"Should the pyQt based gui be compiled and installed" OFF )
-option( ERT_USE_OPENMP      	"Use OpenMP - currently only in EclGrid" OFF)
-option( ERT_DOC             	"Build ERT documantation" OFF)
-option( ERT_BUILD_CXX       	"Build some CXX wrappers" ON)
+
+# try import python module, if success, check its version, store as PY_module
+function(find_python_module_version module)
```
jepebe: Maybe mention that the spelling/casing should be identical to the way the module is imported in Python?

devel/CMakeLists.txt
```diff
@@ -17,17 +17,73 @@ set( ERT_VERSION_MICRO git )
...
+  set(PY_VERSION_ACCESSOR "__version__")
+  set(PY_module_name ${module})
+  if(${module} MATCHES "PyQt4")
+    set(PY_module_name "PyQt4.Qt")
+    set(PY_VERSION_ACCESSOR "PYQT_VERSION_STR")
+  endif(${module} MATCHES "PyQt4")
+  execute_process(COMMAND "python" "-c"
```
jepebe: It would be easier to read this function if you had som additional line breaks...

devel/CMakeLists.txt
jepebe: `find_python_module_version` and `find_python_module` are functions that should be included from the cake directory. There are some examples further down in this file.

PR #1154, opened 2016-07-04
Structs for output of observation names in analysis modules

arielalmendral:
Continuation of #1111

### DIFF
--- a/devel/libanalysis/src/fwd_step_enkf.c
+++ b/devel/libanalysis/src/fwd_step_enkf.c
@@ -27,6 +27,8 @@
 #include <ert/util/matrix.h>
 #include <ert/util/matrix_blas.h>
 #include <ert/util/stepwise.h>
+#include <ert/util/stringlist.h>
+#include <ert/util/double_vector.h>
 
 #include <ert/analysis/fwd_step_enkf.h>
 #include <ert/analysis/fwd_step_log.h>
@@ -34,6 +36,8 @@
 #include <ert/analysis/analysis_module.h>
 #include <ert/analysis/module_data_block.h>
 #include <ert/analysis/module_data_block_vector.h>
+#include <ert/analysis/module_obs_block.h>
+#include <ert/analysis/module_obs_block_vector.h>
 
 
 #define FWD_STEP_ENKF_TYPE_ID 765524
@@ -110,13 +114,11 @@ void fwd_step_enkf_data_free( void * arg ) {
 
 
 static void fwd_step_enkf_write_log_header( fwd_step_enkf_data_type * fwd_step_data, const char * ministep_name, const int nx, const int nd, const int ens_size) {
-  const char * format = "%-15s%-15s%-15s%-15s%-15s%-15s\n";
-  const char * column1 = "Parameter";
-  const char * column2 = "ActiveIndex";
-  const char * column3 = "GlobalIndex";
-  const char * column4 = "NumAttached";
-  const char * column5 = "FinalR2";
-  const char * column6 = "AttachedObs";
+  const char * format = "%-25s%-25s%-25s%-25s\n";
+  const char * column1 = "Parameter(ActiveIndex)";
+  const char * column2 = "GlobalIndex";
+  const char * column3 = "NumAttached";
+  const char * column4 = "AttachedObs(ActiveIndex)[Percentage sensitivity]";
   int nfolds      = fwd_step_data->nfolds;
   double r2_limit = fwd_step_data->r2_limit;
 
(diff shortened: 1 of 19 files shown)


arielalmendral:
Pls wait with reviewing this... WIP.

jepebe:
Just prefix the PR title with WIP:

arielalmendral:
Ready for review, thx. 

I''ll be out of office from Friday, until the beginning of August. Significant required changes will have to wait until I am back.

arielalmendral:
I do not understand this failing test. The test is done on a random vector, so I guess there is a small (but possible) probability of failure within this tested accuracy. I cannot see the relationship with this PR either. 

FAIL: test_stat_quantiles (core.util.test_stat.StatTest)
----------------------------------------------------------------------
  File "/home/travis/build/Ensembles/ert/build/python/tests/core/util/test_stat.py", line 15, in test_stat_quantiles
    self.assertAlmostEqual(quantile_sorted(v, 0.3), 0.3, 2)
AssertionError: 0.29481583971165276 != 0.3 within 2 places

joakim-hove:
> I do not understand this failing test. The test is done on a random vector, so I guess there is a small (but possible) probability of failure within this tested accuracy. I cannot see the relationship with this PR either.

You are absolutely right - there is a low but finite probablity for failure of this test; we should fix the seed.

devel/libanalysis/src/module_obs_block_vector.c
```diff
@@ -0,0 +1,76 @@
...
+  module_obs_block_type * module_obs_block = NULL;
+  for (int block_nr = 0; block_nr < module_obs_block_vector_get_size( module_obs_block_vector ); block_nr++) {
+    module_obs_block_type * module_obs_block =  module_obs_block_vector_iget_module_obs_block (module_obs_block_vector, block_nr);
+    int row_start =  module_obs_block_get_row_start(module_obs_block);
+    int row_end =  module_obs_block_get_row_end(module_obs_block);
+    if (global_index >= row_start && global_index < row_end)
+      return module_obs_block;
+  }
+
+  return module_obs_block;
```
pgdr: Maybe this can just be `return NULL` and remove the instantiation before the for-loop?  It's a bit confusing to read as it is now since the variable is shadowed inside the for-loop scope.

pgdr:
This looks good and seems ready to merge after addressing my question above.  However, I think I want to let someone else ( @jepebe @joakim-hove ) hit the merge-button, unless given permission.

arielalmendral:
Thanks! The question has been addressed.

devel/libanalysis/src/fwd_step_enkf.c
```diff
@@ -140,40 +143,84 @@ static void fwd_step_enkf_write_log_header( fwd_step_enkf_data_type * fwd_step_d
...
+          obs_active_index = active_indices[local_index];
+        break;
+      }
+      local_index ++;
+    }
 
-    printf("%d ",ivar);
+    char * obs_list_entry = util_alloc_sprintf(format1 , obs_key, obs_active_index,var_beta_percent);
+    stringlist_append_copy(obs_list, obs_list_entry);
+    double_vector_append(r_list, var_beta_percent);
```
joakim-hove: Missing: `free( obs_list_entry )`

devel/libenkf/src/enkf_main.c
```diff
@@ -770,17 +768,50 @@ static int enkf_main_serialize_dataset( const ensemble_config_type * ens_config
...
-      const module_data_block_type * data_block = module_data_block_alloc( key, active_list_get_active(active_list),  row_offset[ikw], active_size[ikw] );
-      module_data_block_vector_add_data_block(module_data_block_vector, data_block);
     }
   }
   matrix_shrink_header( A , current_row , ens_size );
   stringlist_free( update_keys );
   return matrix_get_rows( A );
 }
 
+static void enkf_main_init_module_info( const obs_data_type * obs_data,
```
joakim-hove: This function does not seem to be in use at all? If possible I would prefer if [RAII](https://en.wikipedia.org/wiki/Resource_Acquisition_Is_Initialization) could be used - i.e.
```C
module_info_type * enkf_main_alloc_module_info( ... )  {
    // Create and initialize the module_info instance.
} 
```

devel/libanalysis/src/module_obs_block_vector.c
```diff
@@ -0,0 +1,75 @@
...
+  return module_obs_block_vector;
+}
+
+
+void module_obs_block_vector_free( module_obs_block_vector_type * module_obs_block_vector ) {
+  vector_free( module_obs_block_vector->obs_block_vector );
+  free( module_obs_block_vector );
+}
+
+void module_obs_block_vector_add_obs_block( module_obs_block_vector_type * module_obs_block_vector , const module_obs_block_type * obs_block) {
```
joakim-hove: The argument will be mutable in the vector - better change the signature to take a mutable argument.

devel/libanalysis/src/module_obs_block_vector.c
```diff
@@ -0,0 +1,75 @@
...
+  vector_append_owned_ref(module_obs_block_vector->obs_block_vector, obs_block , module_obs_block_free__);
+}
+
+
+module_obs_block_type * module_obs_block_vector_iget_module_obs_block(const module_obs_block_vector_type * module_obs_block_vector, int block_index){
+ return vector_iget(module_obs_block_vector->obs_block_vector, block_index);
+}
+
+const module_obs_block_type * module_obs_block_vector_search_module_obs_block(const module_obs_block_vector_type * module_obs_block_vector, int global_index){
+  for (int block_nr = 0; block_nr < module_obs_block_vector_get_size( module_obs_block_vector ); block_nr++) {
```
joakim-hove: When the loop can - and typically will - break early I prefer a `while( )` loop:

```C
module_obs_block_type * module_obs_block = NULL;
int block_nr = 0;
while (true) {
    if (block_nr >= modul_obs_block_vector_get_size( module_obs_block_vector );
       break;
 
     if () 
        break;
    
    block_nr++;
}
return module_obs_block;
```

joakim-hove:
OK; this looks fine. Will merge when the minor comments have been addressed - *and* some basic tests; at the very least instantiation and destruction of objects.

arielalmendral:
Thanks for the good comments. Fixed accordingly and added some basic tests.

PR #1158, opened 2016-07-06
Ert 1059 Refactoring of Ert GUI event system

jepebe:
The main goal of this PR is remove the `ErtConnector` concept. Instead of using this class any models requiring an `EnKFMain` instance will find it in the constant `ert_gui.ERT`. This new class is called `ErtNotifier` and keeps track of the current ert instance in addition to being the hub for any Ert change notifications.


pgdr:
Sweet!  Looking forward to try it out!

devel/python/python/ert_gui/ertwidgets/checklist.py
```diff
@@ -0,0 +1,154 @@
+from PyQt4.QtCore import QSize, Qt
+from PyQt4.QtGui import QToolButton, QHBoxLayout, QLabel, QListWidget, QWidget, QVBoxLayout, QListWidgetItem, QMenu, QAction, QAbstractItemView, QListView
```
pgdr: How long is too long?  I know you're not a fan of 80 column restriction, but 160 is maybe a bit much?

devel/python/python/ert_gui/ertwidgets/models/selectable_list_model.py
```diff
@@ -0,0 +1,46 @@
+from PyQt4.QtCore import QObject, pyqtSignal
+
+
+class SelectableListModel(QObject):
```
pgdr: If this inherits `QAbstractListModel` it can be plugged directly into any of Qt's `ListView`.

devel/python/python/ert_gui/ertwidgets/checklist.py
```diff
@@ -0,0 +1,154 @@
...
+
+from ert_gui.ertwidgets import addHelpToWidget
+from ert_gui.widgets.search_box import SearchBox
+from ert_gui.widgets.util import resourceIcon
+
+
+class CheckList(QWidget):
```
pgdr: This could essentially be a `QListView`.  See [QListView.setModel](http://pyqt.sourceforge.net/Docs/PyQt4/qabstractitemview.html#setModel).

devel/python/python/ert_gui/ertwidgets/caselist.py
```diff
@@ -0,0 +1,65 @@
...
+            self._list.setSelectionMode(QAbstractItemView.NoSelection)
+
+
+    def addItem(self):
+        dialog = ValidatedDialog("New case", "Enter name of new case:", getAllCases())
+        new_case_name = dialog.showAndTell()
+        if not new_case_name == "":
+            selectOrCreateNewCase(new_case_name)
+
+    def removeItem(self):
```
pgdr: If removing is disabled, then we don't need to display a message to the user.

devel/python/python/ert_gui/ertwidgets/models/selectable_list_model.py
```diff
@@ -0,0 +1,46 @@
...
+        self._items = items
+
+    def getList(self):
+        return self._items
+
+    def isValueSelected(self, value):
+        if self._selection.has_key(value):
+            return self._selection[value]
+        else:
+            return True
```
pgdr: Should this not be `False`, or are all values selected by default?

pgdr: Can't you use `self._selection.get(value, True)`, which returns the second argument if `value not in self._selection`?

jepebe: Done!

devel/python/python/ert_gui/ertwidgets/caselist.py
jepebe: The class used to be generic. Keeping it around if it will be generic again...

devel/python/python/ert_gui/ertwidgets/checklist.py
jepebe: Done.

devel/python/python/ert_gui/ertwidgets/models/selectable_list_model.py
jepebe: I agree. My goal in this round, however, is removing dependency on ErtConnector. So this is a future task.

devel/python/python/ert_gui/ertwidgets/checklist.py
jepebe: See SelectableListModel comment

joakim-hove:
OK;

this seems like 'good-old-maintenance' - very good indeed. Could you please briefly describe the principles behind the changes in the PR description? I guess this really marks the start of 2.1 development - that is good.

Could you raise the version number in the master branch to `2.1.git`.

jokva:
Summer cleaning is going well!

joakim-hove:
I have just launched gert –v gui-test and got this message in the console. Also, the WPRO option is not available in the Plug-ins

```txt
bash> gert -v gui-test config.ert
The script '/project/res/etc/ERT/workflows/gui-plugins/scripts/wpro.py' caused an error during load:
ImportError: No module named models.mixins.connectorless
```

joakim-hove:
When I tried running MDA I got:

```Python
  File "/project/res/x86_64_RH_6/share/ert/release/gui-test/python/ert_gui/simulation/simulation_panel.py", line 102, in runSimulation
    arguments = self.getSimulationArguments()
  File "/project/res/x86_64_RH_6/share/ert/release/gui-test/python/ert_gui/simulation/simulation_panel.py", line 92, in getSimulationArguments
    return simulation_widget.getSimulationArguments()
  File "/project/res/x86_64_RH_6/share/ert/release/gui-test/python/ert_gui/simulation/multiple_data_assimilation_panel.py", line 106, in getSimulationArguments
    "target_case": self._target_case_model.getValue(),
AttributeError: 'MultipleDataAssimilationPanel' object has no attribute '_target_case_model'
```

@jepebe: Line 106 in multiple_data_assimilation_panel.py must be changed to:
"target_case": self._target_case_format_model.getValue(),

devel/python/python/ert_gui/about_dialog.py
```diff
@@ -1,6 +1,6 @@
 #  Copyright (C) 2014  Statoil ASA, Norway.
 #
-#  The file 'closable_dialog.py' is part of ERT - Ensemble based Reservoir Tool.
+#  The file 'closabledialog.py' is part of ERT - Ensemble based Reservoir Tool.
```
pgdr: It's actually `about_dialog`, but "this file" works better in my opinion.

arielalmendral:
I tried to run ES-MDA and I got this error: 

    target_case_name = target_case_format.getValue() % 0
AttributeError: 'str' object has no attribute 'getValue'

I remove .getValue() in all places it appears and it worked...

jepebe:
Thank you! Your solution is the correct fix.

arielalmendral:
Fix in #1190

arielalmendral:
I am getting this error when changing the weights in ES-MDA:

Line 173 of multiple_data_assimilation.py:

    elements = [element.strip() for element in elements if not element.strip() == ""]
AttributeError: 'QString' object has no attribute 'strip'

arielalmendral:
Converting to string from Qstring fixes this:

elements = [str(element).strip() for element in elements if not str(element).strip() == ""]

pgdr:
I think the original `.getValue()` returned a string value, but now that that is removed, you have to cast to string somewhere else.  If you want to use `Qstring`, you must use `trimmed()` and not `strip()`.

arielalmendral:
This error happens when parsing the weights, only when you press run. The .getValue change was related to parsing the target file system.

jepebe:
The best solution is to change `setWeights()` to `str` the input argument.

arielalmendral:
As in #1191 ?

PR #1159, opened 2016-07-10
Isolate ert.cwrap

pgdr:
Moved cwrap to own module.  [OPM/opm-parser#875](https://github.com/OPM/opm-parser/pull/875) depends on this.

* As per suggestion by @joakim-hove `cwrap` has been isolated and now works as a module of its own.
* If you happen to have the `cwrap` module in your Python path, we do not roll our own
* update all the places we use `ert.cwrap` to use `cwrap`
* use new pycmake functionality from [Statoil/pycmake](https://github.com/Statoil/pycmake)

Obs: Many files have been touched in this PR; most of them a result of
```bash
find . -name *.py -exec sed -i "s/ert.cwrap/cwrap/g" {} \;
```


devel/python/CMakeLists.txt
```diff
@@ -7,22 +7,23 @@ if (NOT DEFINED PYTHON_EXECUTABLE)
    return()
 endif()
 
+python_module(cwrap REQUIRED )
 
-python_module(numpy 1.7.1)
+python_module(numpy REQUIRED 1.7.1)
```
joakim-hove: Not 100% clear to me what is the best way to do this, but the requirements are:

1. It should be possible and valid to build a C only target.
2. The python target `ert.ecl` *requires* `numpy` and  `cwrap`.

To get graceful degradation in the case where the user has requested Python bindings, but does not have `numpy`or `cwrap` we can not have `REQUIRED` on `numpy` and `cwrap`.

joakim-hove:
This is *very cool* - I am quite scared to "Rock the boat" with respect to opm; which should "Just work" - apart from that I will merge as soon as it is green in Travis.

jokva:
opm boxes with already-installed ert shouldn't notice this for a while, no? They don't really use the python things and nothing has changed libecl-wise except for finding and declaring python, so at least for now it should be fine.

joakim-hove:
> opm boxes with already-installed ert shouldn't notice this for a while, no?

Well - there is a fine-print to this; 

1. The *all-default* ert build, which is typically used in opm setting, will default to build Python wrappers. If we set `numpy` and `cwrap` as `REQUIRED` this *all-default* ert build can fail.
  
2. Both `opm-parser` and the integration tests in opm-simulators will *look for* the Python bindings; but they will work happily without them.

So - the important point for me is a that the *all-default* build should not fail in an environment which is otherwise good enough for the rest of opm.

jokva:
Good point. I also didn't consider CI, because issues won't really show up until someone does a new build of ert for their project. Which for developers shouldn't happen too often, but in CI it happens all the time.

devel/python/CMakeLists.txt
pgdr: I agree, the addition of the `REQUIRED` keyword was because the `python_module` changed.  The two lines are identical.

That said, I will make it `OPTIONAL`.  The user will get a warning if `numpy` is not installed, but everything (incl. Python) will be built.

joakim-hove: > The user will get a warning if numpy is not installed, but everything (incl. Python) will be built.

Well - then you build a Python distribution which is not functional - I would prefer if Python was just skipped if the required libraries (i.e. `numpy` in this case) was not present.

jokva:
Just bringing it up - would it be reasonable to host crwap somewhere, then use git submodule or something to "provide our own" rather than just embedding the full source tree? I'm pretty sure we can use the submodule from the build system anyway, effectively getting this behaviour, but without maintaining a compatible copy in-tree.

pgdr:
@jokva Yes, that would be reasonable, and indeed that's where we are going.  We already have [Statoil/cwrap](https://github.com/Statoil/cwrap) which is intended as a fully functional stand-alone python module.

Exactly how it should be shipped, I don't know.  But a nice goal is to get it into pip.

joakim-hove:
> Yes, that would be reasonable, and indeed that's where we are going

I agree on the direction here - but my oh my - I would like to go proceed carefully on this.

pgdr:
This PR is ready for review now.

pgdr:
@joakim-hove sure we'll proceed carefully.  This PR is simply isolating the necessary parts.  The next step is replacing the now-isolate `cwrap` library with the new one in Statoil/cwrap.  These two should be minor.

Whether or not we decide on putting cwrap in pip or how we will ship it in the future doesn't really matter.  For all I care, we can continue having it as it is today (but CNamespace and CFILE and the rest updated according to Statoil/cwrap).

devel/python/CMakeLists.txt
```diff
@@ -1,28 +1,40 @@
...
+
+# numpy is needed for Python
+
+set( numpy_target_version '1.7.1')
+
+python_module( numpy OPTIONAL )
 if (NOT DEFINED PY_numpy)
-   message("numpy module not found - Python wrappers not enabled")
-   return()
+  message(WARNING "numpy module not found.  Python wrappers not enabled.")
```
joakim-hove: I agree this *is* a warning - semantically; but the warnings are very ugly - with a cmake backtrace we are not really interested in.

devel/python/CMakeLists.txt
```diff
@@ -1,28 +1,40 @@
 include(cmake/find_python_module.cmake)
-include(cmake/python_module_versions.cmake) # finds version
+include(cmake/pycmake.cmake) # finds version
```
joakim-hove: To facilitate simpler use of the `pycmake` as an external module we could:

1. Put it somewhere "funny" - e.g. `cmake/Modules/pycmake.cmake`.
2. Moidfy the `cmake` search path:

```txt
list( APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/Modules")
```
3. Use an `include( pycmake )` - without prefix path and without extension.

joakim-hove: It seems varieties of `CamelCAPS` is all the rage when it comes to [cmake modules](https://cmake.org/cmake/help/v3.0/manual/cmake-modules.7.html) - so maybe this should be `PyCMake`?

joakim-hove:
Good if you can prepare a PR here as well: https://github.com/OPM/opm-parser/tree/master/opm/parser/eclipse/python/python/opm - the `s/ert.cwrap/cwrap/g` should take you a *very* long way.

joakim-hove:
OK - when you addressed the minor comments; or alternatively convinced me otherwise - I am ready to take a deep breath and merge this!

devel/python/CMakeLists.txt
pgdr: Maybe `FindPythonModule` or something is better.  Any suggestions?

joakim-hove: +1

pgdr:
I'll make the parser PR.

There is no rush merging this PR!  I'll rename the PR to "WIP" so nobody merges it before we are comfortable merging it.

devel/python/CMakeLists.txt
pgdr: Okey, then I'll not warn, I'll simply output a status message (no backtrace).

joakim-hove:
OK;  

there is no slightly serious issue with this approach which we must fix. In the current implementation the `ert/__init__.py` does some monkey-patching of the `ert/cwrap/clib.py` module with the path to the shared libraries. The current approach works to some extent - but it must be reworked and improved before we can make a stand-alone `cwrap` package.

jepebe:
I agree and I think that removing the **old** variants of wrapping is a good first step. Then we only need to solve `cfile` and `clib` issues.

pgdr:
This is ready for review.

pgdr:
Added cwrap as an alias in ert module for backwards compatibility.

pgdr:
It seems that the simple forward alias in ert wasn't successful.  One can import `ert.cwrap` but I'm not exactly sure what that _means_ ...  One is not, on the other hand, able to use `ert.cwrap` beyond importing it.

joakim-hove:
> It seems that the simple forward alias in ert wasn't successful. One can import ert.cwrap but I'm not exactly sure what that means ... One is not, on the other hand, able to use ert.cwrap beyond importing it.

OK - then the remaining steps I see are:

1. We just remove: 19a281ea0fec060033215b5878a45e265cddd1e9 from the PR.
2. We add a PR to opm-common with `FindCwrap` - something resembling the current [FindERTPython](https://github.com/OPM/opm-common/blob/master/cmake/Modules/FindERTPython.cmake)
3. The PR in opm-parser: https://github.com/OPM/opm-parser/pull/875 is updated.

When 2) has been merged we merge 1) and 3) together.

PR #1170, opened 2016-08-05
Inlining GET_INDEX and speed up vector iget

arielalmendral:
Discussion PR... I'd appreciate your comments on this. 

According to Intel Vtune amplifier, there are speed improvements that can achieved by optimizing GET_INDEX (matrix.c) and iget function from vector_template.h.

Questions: 

- Is NDEBUG the flag assigned from cmake in ERT? The assert in iget should be done only in Debug mode. 
- Should ..iget be inlined also? 
- Do we want to force inline of GET_INDEX always (as here) or only in Release mode?

### DIFF
--- a/devel/libert_util/include/ert/util/matrix.h
+++ b/devel/libert_util/include/ert/util/matrix.h
@@ -31,12 +31,42 @@
 #include <ert/util/thread_pool.h>
 #endif
 
+#ifdef _MSC_VER
+#define __forceinline inline
+#elif __GNUC__
+/* Also clang defines the __GNUC__ symbol */
+#define __forceinline inline __attribute__((always_inline))
+#endif
+
+
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+struct matrix_struct {
+  UTIL_TYPE_ID_DECLARATION;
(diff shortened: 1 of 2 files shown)


joakim-hove:
The matrix inline is fine. For the vector I am a bit uncertain - the check does catch quite a number of bugs.  Performance critical code should ask the vector for a raw pointer and access that with [.]?

pgdr:
(meta) I restarted the Travis.gcc job after it failed for some strange reason.  Now it passes.  However, Travis.clang still fails, but that's also due to some strange reasons.

However, I think the AppVeyor failure is an issue we need to address.  It seems to point at something in `matrix.h:61`.

```c
static inline __attribute__((always_inline)) size_t GET_INDEX( const matrix_type * m ,
                                                               size_t i , size_t j)
{
    return m->row_stride *i + m->column_stride *j;
}
```

arielalmendral:
It seems that VC uses a different way of forcing inlining. Will force it only for GCC.

For the vector iget, it is fine to do the operations on the vector's raw data within performance-critical loops. Still think however that the release version shouldn't have the assert - but understand that it depends on the type of stack trace you want reported.

PR #1171, opened 2016-08-08
Fixed bug with bind in EclKw._assert_binary()

joakim-hove:


jepebe:
I think the "real" bug is that the `__IMUL__` function uses `_assert_binary` incorrectly. It explicitly sends `self`. Lines 524, 555, 670 uses `_assert_binary` only two of them are correct.

joakim-hove:
> I think the "real" bug is that the __IMUL__ function uses _assert_binary incorrectly.

Yes. Have fixed this - and updated test sightly.

PR #1177, opened 2016-08-10
Version jobs

joakim-hove:


devel/libenkf/tests/enkf_workflow_job_test_version.c
```diff
@@ -0,0 +1,80 @@
...
+  return 1;
+}
+
+
+int version_get_minor_ert_version( ) {
+  return 2;
+}
+
+
+const char* version_get_micro_ert_version( ) {
```
pgdr: micro=patch?

I think it's common to operate with `major.minor[.patch[.tweak]]` name scheme.

joakim-hove: That is maybe a good idea, but the `micro` name is already in use - also externally, the code you have commented on here is only the *for-testing-override* of the methods.

PR #1178, opened 2016-08-11
change directory to standard repository layout

andlaus:
this patch is large but rather trivial: it moves everything in the
'devel' subdirectory one level up. this implies that external build
scripts must be (slightly) adapted...


andlaus:
Pushed a patch which attempts to fix the CI systems.

joakim-hove:
I will merge a variety of this; but I would prefer:

1. The `devel/` directory is renamed `src/`.
2. The `src/CMakeLists.txt` is moved to the root directory, alternatively the root directory can get a new `CMakeLists.txt` file which essentially only contains `include(src)` and a project header.

What do you think?

andlaus:
I started on this before I read your answer on the OPM mailing list. I'll do as you like, but I've got a question:

> The devel/ directory is renamed src/.

don't you think that a `src` subdirectory would be just noise? (e.g., on my machine the sources of all projects I need to compile sit in the directory `~/src`, so the source of ERT would end up in `~/src/ert/src` which looks slightly weird IMO.)

joakim-hove:
> don't you think that a src subdirectory would be just noise?

Maybe - but there is some other crap there already; so I think it is clean to have a `src/` directory.

andlaus:
> Maybe - but there is some other crap there already; so I think it is clean to have a src/ directory. 

ok, but there are also several `src` directories in subdirectories already:

```
# find -name "src"
./libert_utilxx/src
./libgeometry/src
./libecl/src
./libanalysis/src
./libeclxx/src
./libert_util/src
./libjob_queue/src
./libenkf/src
./libecl_well/src
./libconfig/src
./librms/src
./libsched/src
```

(and also some 'src' directories get created in the build directory, which may confuse people.)

joakim-hove:
OK - I will merge this; or a functionally equivalent alternative soon - but since it touches so many files, and also more than one repo must be updated, so some coordination is required.

andlaus:
> OK - I will merge this; or a functionally equivalent alternative soon - but since it touches so many files, and also more than one repo must be updated, so some coordination is required.

cool. feel free to merge all other open PRs first. the conflicts with this one here are probably easier to deal with than for the others. (I'll do it for you once you tell me.)

jepebe:
Thank you! I was going to suggest that it wait until after PR #1158 which has a couple of new files and a file move or two...

andlaus:
ok, this has been superseeded by #1196. closing.

PR #1179, opened 2016-08-11
Fix bug missing assignment of analysis module in ES-MDA

arielalmendral:


devel/python/python/ert_gui/models/connectors/run/multiple_data_assimilation.py
```diff
@@ -50,8 +50,10 @@ def runSimulations(self):
         iteration_count = len(weights)
 
         print("Running MDA ES for %s  iterations\t%s" % (iteration_count, ", ".join(str(weight) for weight in weights)))
-        weights = self.normalizeWeights(weights)
 
+        self.setAnalysisModule()
```
joakim-hove: I do not understand this:

1. An analysis module has already been assigned - either the default module, or something from the configuration file.
2. I would assume the `setAnalysisModule()` method would require an argument?

arielalmendral:
Set module does not need an argument, as currently implemented. It picks the "current choice" from the UI: 
```
def setAnalysisModule(self):
 module_name = AnalysisModuleModel().getCurrentChoice()
...
```
ES-MDA has opened the possibility of picking different modules now.

You may a see similar assignment for the other two schemes: ensemble_smoother and iterated_ensemble_smoother

jepebe:
I think this functionality was addressed in the mega PR #1158

arielalmendral:
OK, closing this one then.

PR #1182, opened 2016-08-11
Fixing several warnings, const/unconst and perm_vector related

arielalmendral:


devel/libenkf/src/enkf_main.c
```diff
@@ -1320,7 +1320,7 @@ bool enkf_main_smoother_update(enkf_main_type * enkf_main , enkf_fs_type * sourc
 
 
 
-
+#ifdef QUEUE_FAILURE_CODE
```
joakim-hove: What is this `#ifdef`?

arielalmendral:
The code is not in use...

joakim-hove:
> The code is not in use...

OK - then we remove it.

joakim-hove:
I'll merge when the code which had been `#ifdef`'ed out is removed.

arielalmendral:
Unused code removed.

joakim-hove:
Thank you - in she goes.

PR #1186, opened 2016-08-12
Added documentation of UMASK.

joakim-hove:


devel/docs/user/keywords/index.rst
```diff
@@ -1862,3 +1862,45 @@ The two keywords SETENV and UPDATE_PATH can be used to manipulate the Unix envir
...
+
+        To determine the initial permissions on newly created files
+        start with the initial permissions `-rw-rw-rw-` (octal 0666)
+        for files and `-rwxrwxrwx` (octal 0777) for directories, and
+        then *~subtract* the current umask setting. So if you wish the
+        newly created files to have permissions `-rw-r-----` you need
+        to subtract write permissions for group and read and write
+        permissions for others - corresponding to `umask
+        0026`. 
+
```
jepebe: Would it make sense to have this list here?

Octal value : Permission
0 : read, write and execute
1 : read and write
2 : read and execute
3 : read only
4 : write and execute
5 : write only
6 : execute only
7 : no permissions

joakim-hove:
> Would it make sense to have this list here?

Added some documentation of octal permissions; but in the "permission space" and not in the inverted "umask space".

jepebe:
Great!

PR #1188, opened 2016-08-15
Added EclKW.numpyView( ) and EclKW.numpyCopy( )

joakim-hove:


devel/python/python/ert/ecl/ecl_kw.py
```diff
@@ -1053,17 +1053,47 @@ def __str__(self):
         Return string representation - see method str().
         """
         return self.str( width = 5 , max_lines = 10)
-    
 
 
+    def numpyView(self):
+        """
+        Will return a numpy view to the underlying data.
```
jepebe: Explain difference between Copy and View

PR #1194, opened 2016-08-18
Plugin doc

joakim-hove:


devel/docs/user/workflows/index.rst
```diff
@@ -7,16 +7,22 @@ Contents
...
+*workflows*. Using workflows way you can automate cumbersome normal
+ERT processes, and also invoke external programs. The workflows are
+run serially on the workstation actually running ERT, and should not
+be used for computationally heavy tasks.
 
+In addition to workflows ERT has a system for *plugins*, a plugin is
+quite similar to a workflow job, but the plugin appears in a gui
+dropdown, and can even have a gui of it's own. A plugin is written in
+Python and has *full access* to ERT internals. Plugins can only be
+invoked from the gui.
```
jepebe: They can also be run from ertshell.

devel/docs/user/workflows/plugins.rst
```diff
@@ -0,0 +1,45 @@
...
+file :code:`script/ens_size.py` then it is installed in ert as:
+
+   INTERNAL  TRUE                 -- The job will call an internal function of the current running ERT instance.
+   SCRIPT sripts/ens_size.py      -- An existing Python script
+
+ERT will detect that the class inherits from :code:`ErtPlugin`, and
+the functionality will be available as :code:`Size plugin` in the plugin menu.
+
+That is not demonstrated here, but plugins can have a GUI of their own.
+
```
jepebe: Maybe point to `csv_export` which is a plugin and has a GUI.

joakim-hove:
> Maybe point to csv_export which is a plugin and has a GUI.

Have updated the documentation with an inline of the "CSV export" plugin.

jepebe:
Nice!

PR #1196, opened 2016-08-20
Move cmake root

joakim-hove:
This is for all practical purposes the work of @andlaus( #1178 ) - but rebasing the monster commit was a bit troublesome, so I recreated that and cherry-picked the remaining commits.


pgdr:
Hats off to @joakim-hove and @andlaus.

pgdr:
Perhaps we should just merge this now and deal with any potential trouble later?

joakim-hove:
> Perhaps we should just merge this now and deal with any potential trouble later?

The plan was to do it tomorrow - just let this: https://github.com/OPM/opm-common/pull/167 linger a short while.

PR #1199, opened 2016-08-22
Ecl file set parent

joakim-hove:
The most important commit here is: c56d0d0 which fixes a real bug with dangling EclKW instances when the parent EclFile goes out of scope. This worked before :-(

Tried some a slightly more magic approach based on Metaclasses - but that failed.


python/python/ert/ecl/ecl_file.py
```diff
@@ -358,6 +358,15 @@ def select_last_restart( self ):
             return False
 
 
+    def __iget__(self , index):
```
jepebe: Remove two last underscores

joakim-hove: Update pushed.

PR #1204, opened 2016-08-23
Add nnc index to the information exported by ecl_nnc_export().

atgeirr:
This is necessary and sufficient to recover original nnc ordering for the purpose of reading connection fluxes for example, in the absence of local grid refinement.

For the LGR case more information (such as keyword) will be required.

### DIFF
--- a/libecl/include/ert/ecl/ecl_nnc_export.h
+++ b/libecl/include/ert/ecl/ecl_nnc_export.h
@@ -38,7 +38,7 @@ typedef struct {
   int grid_nr2;
   int global_index2;
 
-  int nnc_index;
+  int input_index; /* corresponds to the input ordering of this nnc */
 
   double trans;
 } ecl_nnc_type;
--- a/libecl/include/ert/ecl/ecl_nnc_export.h
+++ b/libecl/include/ert/ecl/ecl_nnc_export.h
@@ -38,6 +38,8 @@ typedef struct {
   int grid_nr2;
   int global_index2;
 
+  int nnc_index;
+
   double trans;
 } ecl_nnc_type;
 
--- a/libecl/src/ecl_nnc_export.c
+++ b/libecl/src/ecl_nnc_export.c
@@ -64,9 +64,9 @@ static int  ecl_nnc_export__( const ecl_grid_type * grid , int lgr_index1 , cons
 
         for (index2 = 0; index2 < nnc_vector_get_size( nnc_vector ); index2++) {
           nnc.global_index2 = int_vector_iget( grid2_index_list , index2 );
-          nnc.nnc_index = int_vector_iget( nnc_index_list, index2 );
+          nnc.input_index = int_vector_iget( nnc_index_list, index2 );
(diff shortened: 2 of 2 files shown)


libecl/include/ert/ecl/ecl_nnc_export.h
```diff
@@ -38,6 +38,8 @@ typedef struct {
   int grid_nr2;
   int global_index2;
 
+  int nnc_index;
```
joakim-hove: Mabye add a comment here: "The `ncc_index` corresponds to the input ordering of this nnc."

libecl/src/ecl_nnc_export.c
```diff
@@ -64,6 +64,7 @@ static int  ecl_nnc_export__( const ecl_grid_type * grid , int lgr_index1 , cons
 
         for (index2 = 0; index2 < nnc_vector_get_size( nnc_vector ); index2++) {
           nnc.global_index2 = int_vector_iget( grid2_index_list , index2 );
+          nnc.nnc_index = int_vector_iget( nnc_index_list, index2 );
           if(tran_kw) {
             nnc.trans = ecl_kw_iget_as_double(tran_kw, int_vector_iget(nnc_index_list, index2));
```
joakim-hove: The new variable `ncc.nnc_index` should be used as second argument in the `nnc.trans` assignment:
```C
nnc.trans = ecl_kw_iget_as_double(tran_kw, nnc.nnc_index);
```

joakim-hove:
> For the LGR case more information (such as keyword) will be required.

There is a quite complex machinery in place to pick up the right keyword in the case of `TRAN???` for lgr related transmissibilities - maybe the same machinery can be used/extended/improved to also handle the flux keywords?

libecl/include/ert/ecl/ecl_nnc_export.h
atgeirr: Perhaps, I chose the name nnc_index, as that was the name used throughout the grid processing code, as well as in the name of the list (nnc_index_list). Your choice: if you want me to change it I will.

libecl/src/ecl_nnc_export.c
atgeirr: Will do.

libecl/include/ert/ecl/ecl_nnc_export.h
joakim-hove: Yes - I know that name was already used, so in that sense your choice was the natural one, I would still prefer the change though.

atgeirr:
I have updated the PR as requested.

> There is a quite complex machinery in place to pick up the right keyword in the case of TRAN??? for lgr related transmissibilities - maybe the same machinery can be used/extended/improved to also handle the flux keywords?

That sounds like something we should look into at least. If the ordering of nnc keywords in EGRID can be different from that of the transmissibility keywords in EGRID or the fluxes in the restart file we are going to need to store more context than if we can assume they are in some canonical order.

PR #1207, opened 2016-08-24
Ecl grid oom

joakim-hove:
Fix bug in ecl_kw_grdecl loader for large files - reported by Resinsight.

### DIFF
--- a/libecl/src/ecl_grid.c
+++ b/libecl/src/ecl_grid.c
@@ -1435,19 +1435,23 @@ static void ecl_grid_taint_cells( ecl_grid_type * ecl_grid ) {
 
 
 static void ecl_grid_free_cells( ecl_grid_type * grid ) {
+  if (!grid->cells)
+    return;
 
-  int i;
-  for (i=0; i < grid->size; i++) {
+  for (int i=0; i < grid->size; i++) {
     ecl_cell_type * cell = ecl_grid_get_cell( grid , i );
     if (cell->nnc_info)
       nnc_info_free(cell->nnc_info);
   }
-
   free( grid->cells );
+
 }
 
-static void ecl_grid_alloc_cells( ecl_grid_type * grid , bool init_valid) {
-  grid->cells           = util_calloc(grid->size , sizeof * grid->cells );
+static bool ecl_grid_alloc_cells( ecl_grid_type * grid , bool init_valid) {
+  grid->cells = malloc(grid->size * sizeof * grid->cells );
+  if (!grid->cells)
+    return false;
+
   {
     ecl_cell_type * cell0 = ecl_grid_get_cell( grid , 0 );
     ecl_cell_init( cell0 , init_valid );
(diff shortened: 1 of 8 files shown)


libecl/src/ecl_grid.c
```diff
@@ -1431,20 +1431,21 @@ static void ecl_grid_taint_cells( ecl_grid_type * ecl_grid ) {
 
 
 static void ecl_grid_free_cells( ecl_grid_type * grid ) {
+  if (grid->cells) {
```
flikka: What about rather flipping this around, and just return early? I think this makes it easier to read.

libecl/src/ecl_grid.c
```diff
@@ -2202,19 +2211,20 @@ static ecl_grid_type * ecl_grid_alloc_GRDECL_data__(ecl_grid_type * global_grid
                                                     int lgr_nr) {
 
   ecl_grid_type * ecl_grid = ecl_grid_alloc_empty(global_grid , dualp_flag , nx,ny,nz,lgr_nr,true);
+  if (ecl_grid) {
```
flikka: What about rather flipping this around, and just return early? I think this makes it easier to read.

libecl/src/ecl_grid.c
```diff
@@ -2711,19 +2727,22 @@ static ecl_grid_type * ecl_grid_alloc_GRID_data__(ecl_grid_type * global_grid ,
     nz = nz / 2;
   {
     ecl_grid_type * grid = ecl_grid_alloc_empty( global_grid , dualp_flag , nx , ny , nz , grid_nr, false);
+    if (grid) {
```
flikka: ... and so on and so forth...

joakim-hove:
Hmmm; this one is tricky. I have updated some of the callsites according to suggestion, but the callsites of this type:
```C
ptr_type * ptr = ptr_alloc( ... );
if (ptr) {
   // Initialize *ptr object
}
return ptr;
```
where the ptr is returned unconditionally - either as NULL or as a pointer to

flikka:
Personally, and honestly, but humbly, I rather find this more readable:

```C
ptr_type * ptr = ptr_alloc( ... );
if (!ptr) {
   return NULL;
}
// Initialize *ptr object
return ptr;
```

I think the reason is that it makes the "non-happy" path more obvious and earlier, and that the fact that the initialization of the objects takes "many lines of code", and thus "hides" (further down) how the non-happy path is dealt with.

But I won't fight, argue or cry - just throwing out my 5 cents :-)

joakim-hove:
Really no big deal - but I guess my take on it is that 
```C
ptr_type * ptr = ptr_alloc( ... );
if (ptr) {
   // Initialize *ptr object
}
return ptr;
```
is easier because it *does not have* a sorry-path.

python/tests/core/ecl/test_grid_statoil.py
```diff
@@ -383,3 +383,7 @@ def test_volume_kw(self):
 
                               
         
+    def test_large_case(self):
+        grdecl_file = self.createTestPath("Statoil/ECLIPSE/1.6.0_issueGrdecl/test_aug2016_gridOnly.grdecl")
+        grid = EclGrid.loadFromGrdecl( grdecl_file )
```
flikka: What does this test test? That a too big grid is not causing a crash? Or that it actually loads? No asserts...

PR #1208, opened 2016-08-24
In case we do not want cross-validation

arielalmendral:
1. The variable selection takes time because of the CV-fold validation. These are the changes when no CV is asked for.

2. A better solution than the one in #1205 is to control CV by the flag CROSS_VALIDATION.


arielalmendral:
Thanks!

PR #1209, opened 2016-08-25
Fixed ecl_nnc test after field has been added.

joakim-hove:


joakim-hove:
OK - this is recovers the internal tests. Merging.

PR #1210, opened 2016-08-26
Added reading ANALYSIS_SCALE_DATA

pgdr:
Added reading ANALYSIS_SCALE_DATA flag for STD_ENKF in config file, default to true

### DIFF
--- a/libanalysis/include/ert/analysis/std_enkf.h
+++ b/libanalysis/include/ert/analysis/std_enkf.h
@@ -14,6 +14,7 @@ extern "C" {
 #define  ENKF_TRUNCATION_KEY_      "ENKF_TRUNCATION"
 #define  ENKF_NCOMP_KEY_           "ENKF_NCOMP"
 #define  USE_EE_KEY_               "USE_EE"
+#define  ANALYSIS_SCALE_DATA_KEY_  "ANALYSIS_SCALE_DATA"
 
   typedef struct std_enkf_data_struct std_enkf_data_type;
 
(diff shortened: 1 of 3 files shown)


joakim-hove:
OK - this is a good start; but unfortunately there must be some additional double book keeping between the new bool variable and the variable `long option_flags` , i.e. the `option_flags`must still be maintained correctly. That is because the calling scope does a get of the complete option value and checks the scaling bit externally: https://github.com/Ensembles/ert/blob/master/libenkf/src/enkf_main.c#L1027

pgdr:
Okay, so when the flag is set, I should also do
```c
data->option_flags &= (value * ANALYSIS_SCALE_DATA)
```
?

joakim-hove:
> Okay, so when the flag is set, I should also do [...]

Yes; I guess that is the natural way to do it.

joakim-hove:
OK - not really clear to me why this fails on Travis; but when that is fixed I will merge.

pgdr:
For some reason
```c
    analysis_table_type * analysis_table = (analysis_table_type *) dlsym( lib_handle , table_name );
```
fails in `libanalysis/src/analysis_module.c:163`.

I don't really know the cause of it.

joakim-hove:
> For some reason
```C
analysis_table_type * analysis_table = (analysis_table_type *) [...]
```
fails.

No - I disagree. The error message on stderr is a normal module load failure message when we actually test loading of a not correctly configured module. 

The test failure was a normal failure due to missing initialization of the scale flag; an alternative implementation bypassing the whole problem here: #1211.

PR #1213, opened 2016-08-29
WIP: Refactor ecl_file to return views.

joakim-hove:


joakim-hove:
Has been merged into #1233

PR #1222, opened 2016-09-02
Reverting allowing for no-cross validation. Cross-validation is needed.

arielalmendral:
Sorry for the back and forth.


joakim-hove:
Hehe - a bit back and forth here yes ;-)

PR #1223, opened 2016-09-04
test_area: sync directory.

joakim-hove:


flikka:
Missing a sync function in the TestAreaContext class in python?

joakim-hove:
> Missing a sync function in the TestAreaContext class in python?

Ehhh yes - that was the correct diagnosis. Problem fixed and PR updated.

flikka:
Unfortunately this did not seem to do the trick...

PR #1224, opened 2016-09-05
Possibility to assign an analysis module to a ministep

arielalmendral:
With this PR we open up the possibility for the user of local configs to change the analysis modules for distinct ministeps. 

The python syntax for this assignment is:

`ministep.setAnalysisModule(ert.analysisConfig(), "FWD_STEP_ENKF")`

The documentation for local configuration will be updated in a separated PR.

### DIFF
--- a/libenkf/src/enkf_main.c
+++ b/libenkf/src/enkf_main.c
@@ -997,7 +997,6 @@ static void enkf_main_analysis_update( enkf_main_type * enkf_main ,
   const int cpu_threads       = 4;
   const int matrix_start_size = 250000;
   thread_pool_type * tp       = thread_pool_alloc( cpu_threads , false );
-  analysis_module_type * module = analysis_config_get_active_module( enkf_main->analysis_config );
   int active_ens_size   = meas_data_get_active_ens_size( forecast );
   int active_size       = obs_data_get_active_size( obs_data );
   matrix_type * X       = matrix_alloc( active_ens_size , active_ens_size );
@@ -1010,6 +1009,9 @@ static void enkf_main_analysis_update( enkf_main_type * enkf_main ,
   matrix_type * localA  = NULL;
   int_vector_type * iens_active_index = bool_vector_alloc_active_index_list(ens_mask , -1);
 
+  analysis_module_type * module = analysis_config_get_active_module( enkf_main->analysis_config );
+  if ( local_ministep_has_analysis_module (ministep))
+    module = local_ministep_get_analysis_module (ministep);
 
   assert_matrix_size(X , "X" , active_ens_size , active_ens_size);
   assert_matrix_size(S , "S" , active_size , active_ens_size);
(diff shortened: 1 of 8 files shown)


joakim-hove:
> ministep.setAnalysisModule(ert.analysisConfig(), "FWD_STEP_ENKF")

I would prefer if the ministep got a `AnalysisModule` proper object; as:

```Python
analyis_module = ert.analysisConfig().get_module("FWD_STEP_ENKF")
ministep.setAnalysisModule( analysis_module )
```

That way we do not add/increase the dependency between ministep and AnalysisConfig,

libenkf/src/enkf_main.c
```diff
@@ -1010,6 +1009,11 @@ static void enkf_main_analysis_update( enkf_main_type * enkf_main ,
   matrix_type * localA  = NULL;
   int_vector_type * iens_active_index = bool_vector_alloc_active_index_list(ens_mask , -1);
 
+  analysis_module_type * module = NULL;
```
joakim-hove: No need to do `if-then-else` here:
```C
analysis_module_type * module = analysis_config_get_active_module( enkf_main->analysis_config );
if ( local_ministep_has_analysis_module (ministep))
     module = local_ministep_get_analysis_module (ministep);
```

libenkf/src/local_ministep.c
```diff
@@ -238,6 +240,19 @@ bool local_ministep_has_data_key(const local_ministep_type * ministep , const ch
...
 
+bool local_ministep_has_analysis_module( const local_ministep_type * ministep){
+  return ministep->analysis_module != NULL;
+}
+
+analysis_module_type* local_ministep_get_analysis_module( const local_ministep_type * ministep ){
+  return ministep->analysis_module;
+}
+
+void local_ministep_set_analysis_module(local_ministep_type * ministep,  analysis_config_type * config , const char * module_name ) {
```
joakim-hove: Should get a module!

joakim-hove:
This is cool!

arielalmendral:
Thanks! Assigning a module directly to a ministep now.

joakim-hove:
Sorry for nagging: I would preger if the module was passed as an (optional) construction argument, and no public setAnalysisModule. What do you think?

arielalmendral:
Assignment in constructor will somewhat restrict the functionality, but I cannot  find an example now where the current public module assignment/switching is needed. Changes coming...

joakim-hove:
> Assignment in constructor will somewhat restrict the functionality

Yes it will - and as long as we can still do what we want I think enforcing *immutability* in this way is a Good Thing&trade;

> Changes coming...

Thank you!

arielalmendral:
Analysis module assigned in constructor and test. Note that the old text-based functionality of local config will not support this addition. 

At some point we should get rid of the old text-text based functionality.  Maybe soon?

joakim-hove:
> At some point we should get rid of the old text-text based functionality. Maybe soon?

Yes - now!

arielalmendral:
Oki, coming next, along with updated doc.

PR #1226, opened 2016-09-12
WIP: ctime defaults to UTC.

joakim-hove:
For discussion. 

With this patch the `ert.util.ctime` class can optionally use `time.gmtime()` instead of `time.localtime()`; default behavior is still to use `time.localtime()`.


jepebe:
I believe the tests illustrates the problem. There is a mismatch between the C timezone (and the timezone stored in files) and the Python timezone. C will always use the local timezone. You can override it with the environment variable TZ.

joakim-hove:
#1229 completely supersedes this.

PR #1228, opened 2016-09-14
Cache cell volumes.

joakim-hove:


libecl/src/ecl_grid.c
```diff
@@ -1230,58 +1231,65 @@ static inline double tetrahedron_volume( tetrahedron_type tet ) {
...
-          tet.p3 = corners[ point2 ];
-          volume += tetrahedron_volume( tet ) / 6;
+	const int point0 = tetrahedron_permutations[ method ][ itet ][ 0 ];
+	const int point1 = tetrahedron_permutations[ method ][ itet ][ 1 ];
+	const int point2 = tetrahedron_permutations[ method ][ itet ][ 2 ];
+	
+	tet.p1 = corners[ point0 ];
+	tet.p2 = corners[ point1 ];
+	tet.p3 = corners[ point2 ];
+	volume += tetrahedron_volume( tet ) / 6;
```
pgdr: And out of nowhere tabs appeared!

Is there a reason you do not `/6` after the looping?

libecl/src/ecl_grid.c
```diff
@@ -694,6 +694,7 @@ static void point_copy_values(point_type * p , const point_type * src) {
 #define CELL_FLAG_TAINTED  4     /* lazy fucking stupid reservoir engineers make invalid grid
                                     cells - for kicks??  must try to keep those cells out of
                                     real-world calculations with some hysteric heuristics.*/
+#define CELL_FLAG_VOLUME 8
```
pgdr: Would it be possible to initialize volume with -1 and skip this flag?

libecl/src/ecl_grid.c
pgdr: Perhaps we could change the name of `tetrahedron_volume` to something else if it doesn't actually compute the volume.

pgdr:
Go ahead and merge once comments are addressed.

libecl/src/ecl_grid.c
joakim-hove: > Would it be possible to initialize volume with -1 and skip this flag?

There are cases (not active now but still) - where the signed volume is interesting.

PR #1230, opened 2016-09-18
Asking explicitly for pyqt4.

joakim-hove:


.travis.yml
```diff
@@ -49,7 +49,8 @@ install:
     - conda update -q conda
     # Useful for debugging any issues with conda
     - conda info -a
-    - conda install pyqt scipy=0.16.1 numpy pandas matplotlib
+    - conda install scipy=0.16.1 numpy pandas matplotlib
+    - pip install pyqt-4
```
jepebe: `conda install` will install the Python libraries in addition to any required binaries while `pip install` assumes Qt binaries already installed.

What you are trying to achieve can be done with this command: `conda install pyqt=4`

joakim-hove: > What you are trying to achieve can be done with this command: conda install pyqt=4

Good - thank you; seems to work now.

PR #1233, opened 2016-09-20
ConfigParser.parse( ): can return invalid content.

joakim-hove:


libert_util/src/util.c
```diff
@@ -3134,6 +3134,35 @@ bool util_is_first_day_in_month_utc( time_t t) {
...
+
+  The validator will check that the two separation characters are the
+  same.
+*/
+
+bool util_sscanf_isodate(const char * date_token , time_t * t) {
+  int day   , month , year;
+  char sep1 , sep2;
+
+  if (sscanf(date_token , "%d%c%d%c%d" , &year , &sep1 , &month , &sep2 , &day) == 5) {
```
jepebe: This should probably be: `%4d%c%2d%c%2d` where the only allowed separators are: `-` and nothing for dates. https://no.wikipedia.org/wiki/ISO_8601

pgdr: "Be conservative in what you do, be liberal in what you accept from others."

Why not accept dots?  Do you foresee any possible problems?

pgdr:
It seems `ert_util_sscan_test` fails:

`ert_util_sscan_test.c:72 => Booleans are different b1:[0]  b2:[1]`

joakim-hove:
Split in #1247, #1248, #1249 and WIP: #1250

PR #1236, opened 2016-09-23
Implemented ecl_rst_file_open_write_seek( )

joakim-hove:
The function ecl_rst_file_open_write_seek( ) can be used to open a
restart file in read-write mode, including a seek to the right seqnum header.

### DIFF
--- a/libecl/src/ecl_rst_file.c
+++ b/libecl/src/ecl_rst_file.c
@@ -1,19 +1,20 @@
+
 /*
-   Copyright (C) 2012  Statoil ASA, Norway. 
-   
-   The file 'ecl_rst_file.c' is part of ERT - Ensemble based Reservoir Tool. 
-    
-   ERT is free software: you can redistribute it and/or modify 
-   it under the terms of the GNU General Public License as published by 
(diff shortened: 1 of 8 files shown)


pgdr:
For some reason this fails because Travis cannot find PyQt4 and ert is set to build gui.

libecl/src/ecl_rst_file.c
```diff
@@ -73,6 +74,67 @@ ecl_rst_file_type * ecl_rst_file_open_read( const char * filename ) {
...
+    while (true) {
+      offset_type current_offset = fortio_ftell( rst_file->fortio );
+
+      if (fortio_read_at_eof(rst_file->fortio)) {
+        if (seqnum_found)
+          target_pos = current_offset;
+        break;
+      }
+
+      if (ecl_kw_fread_header( work_kw , rst_file->fortio)) {
```
jokva: I honestly think this would be easier to follow if you do `if( !fread_header ) break`, since there are three more code paths inside this if that would lead to break, and with 2-space indentation it's quite hard to visually determine which else clause belongs to which condition.

PR #1237, opened 2016-09-23
Ensuring that max_runtime is enforced for WPRO

flikka:
Added some refactoring of the names of the job statuses in queue_driver.h

The main changes, a bit hidden in this renaming forest:
1) Added {get|set}_max_job_duration in queue.py
2) In simulation_context.py, fetch the "max_runtime" from config, and call the set_max_job_duration method.
3) Change didRealizationFail in simulation_context to return !didRealizationSucceed. This is done because the wpro / seba program tries to read results for all jobs that are not failed. Alternatively, one could change wpro/seba to rather call didRealizationSucceed.
4) In job_queue.c, there is a REAL change in the JOB_QUEUE_COMPLETE_STATUS mask, because it used to include JOB_QUEUE_USER_EXIT (now JOB_QUEUE_DO_KILL) - but this is not a final state, JOB_QUEUE_USER_KILLED is (now JOB_QUEUE_IS_KILLED).

Maybe refactoring and code change is a bad idea. I can split it in two commits if that is better...

### DIFF
--- a/python/python/ert/server/simulation_context.py
+++ b/python/python/ert/server/simulation_context.py
@@ -10,8 +10,12 @@ class SimulationContext(object):
         self._ert = ert
         """ :type: ert.enkf.EnKFMain """
         self._size = size
+        
+        max_runtime = ert.analysisConfig().get_max_runtime()
+        job_queue = ert.siteConfig().getJobQueue()
+        job_queue.set_max_job_duration(max_runtime)
 
-        self._queue_manager = JobQueueManager(ert.siteConfig().getJobQueue())
+        self._queue_manager = JobQueueManager(job_queue)
         self._queue_manager.startQueue(size, verbose=verbose)
         self._run_args = {}
         """ :type: dict[int, RunArg] """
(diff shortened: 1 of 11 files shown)


python/python/ert/server/simulation_context.py
```diff
@@ -10,8 +10,12 @@ def __init__(self, ert, size, verbose=False):
         self._ert = ert
         """ :type: ert.enkf.EnKFMain """
         self._size = size
+        max_runtime = ert.analysisConfig().get_max_runtime()
+        print("Max runtime is {}".format(max_runtime))
```
joakim-hove: I guess this `print()` statement is a debugging leftover?

python/python/ert_gui/simulation/models/base_run_model.py
```diff
@@ -178,6 +178,7 @@ def getProgress(self):
             done_count = 0
 
             for state in queue_status:
+		print(state)
```
joakim-hove: Stale debug message?!

joakim-hove:
This has been interesting - it is clear that the codepath followed by WPRO is different from all other simulation modes; maybe it must be like this - but it is somewhat unfortunate.

Apart from the minor `print()` statements lying around I think this should be merged.

flikka:
Good, thanks. Will fix tomorrow!

flikka:
Have removed the prints - I guess it can be merged @joakim-hove

PR #1243, opened 2016-09-28
Added stringlist_join function, similar to pythons str.join

pgdr:


joakim-hove:
Well - this looks good; with test and all - but I guess it is *very* similar to the existing `stringlist_alloc_joined()`?

pgdr:
Yes, I couldn't find that function so I reimplemented it.

I have removed the function, but kept the test.  Merge or close (or review) at will.

PR #1248, opened 2016-09-29
Iso date

joakim-hove:


libert_util/src/util.c
```diff
@@ -3134,6 +3134,28 @@ bool util_is_first_day_in_month_utc( time_t t) {
...
+    *t = -1;
+
+  if (sscanf(date_token , "%d-%d-%d" , &year , &month , &day) == 3) {
+    if (t)
+      *t = util_make_date_utc(day , month , year );
+    return true;
+  }
+
+  return false;
+}
```
pgdr: I must admit that even after reading the implementation, I don't really know what `util_sscanf_isodate` does.

Could you add a small comment somewhere about what the semantic is?

libert_util/tests/ert_util_sscan_test.c
```diff
@@ -65,7 +65,33 @@ void test_sscan_percent() {
...
+    test_assert_time_t_equal( t , expected );
+  else
+    test_assert_time_t_equal( t , -1 );
+}
+
+
+void test_scan_iso_date() {
+  time_t expected = util_make_date_utc(10,  11 , 2011);
+
+  test_date( expected , "2011-11-10", true);
```
pgdr: Could you add at least two more tests, one where `t = -1` and `expected_return=false`, and one where `t=1` and `expected_return=false`?

libert_util/tests/ert_util_sscan_test.c
```diff
@@ -65,7 +65,33 @@ void test_sscan_percent() {
...
+
+  test_assert_false( util_sscanf_isodate( "2017.10.07" , NULL ));
+  test_assert_false( util_sscanf_isodate( "2017-10.7" , NULL ));
+  test_assert_false( util_sscanf_isodate( "2017/10/07" , NULL ));
+
```
pgdr: Maybe also test that `1995-13-01` is not a date, but `1993-12-30` is a date?  So that we know that months are months and days are days?

libconfig/src/config_content_node.c
```diff
@@ -116,6 +116,14 @@ config_item_types config_content_node_iget_type( const config_content_node_type
 }
 
 
+time_t config_content_node_iget_as_isodate(const config_content_node_type * node , int index) {
+  time_t value;
+  config_schema_item_assure_type(node->schema , index , CONFIG_ISODATE);
+  util_sscanf_isodate( config_content_node_iget(node , index) , &value );
+  return value;
```
pgdr: return NULL if sscan is false?

pgdr:
I have some small questions (see review above).  Besides the questions, this is ready to be merged.

libconfig/src/config_content_node.c
joakim-hove: Yes - fixed.

PR #1249, opened 2016-09-29
Config python improvements

joakim-hove:


python/python/ert/config/schema_item.py
```diff
@@ -0,0 +1,46 @@
+from ert.config import ContentTypeEnum , ConfigPrototype
```
pgdr: missing  copyright header

pgdr:
This is a good contribution that I am happy to see merged.

Add the missing copyright header in the schema_item file and this should be good to go.

I would like to see more tests for `schema_item`, but that can be for the future.

python/python/ert/config/schema_item.py
joakim-hove: OK - added.

PR #1250, opened 2016-09-29
Expose ecl_file_view datastructure.

joakim-hove:
The ecl_file structure implements a structured view of an Eclipse file of keywords. An important part of the ecl_file implementation is a index datastructure which keeps track of how the keywords are organized in the file. An important part of the implementation has been that it is possible to create additional views/indexes - which only consider a subset of the keywords. This view datastructure has until know been fully internal, and working with only a subset of keywords is achieved by *modifying* the active index property of the ecl_file structure. With this PR the `ecl_file_view` has become a first-order public object. This is a significant change/extension to the API, but most of the code has been working as an implementation detail for a long time.


pgdr:
Is this ready and in need for reviewing?

joakim-hove:
> Is this ready and in need for reviewing?

Yes - very much so.

python/python/ert/ecl/ecl_file.py
```diff
@@ -293,17 +283,35 @@ def free(self):
...
+        return self.global_view.blockView2( start_kw , stop_kw, start_index )
+
+
+    def restartView( self, seqnum_index = None, report_step = None , sim_time = None , sim_days = None):
+        return self.global_view.restartView( seqnum_index, report_step , sim_time, sim_days )
+    
+
+    
     def select_block( self, kw , kw_index):
+<<<<<<< 542d8044ade962abf7029734d1a29e7e093dc930
```
pgdr: Woops!

libecl/src/ecl_file_view.c
```diff
@@ -0,0 +1,785 @@
...
+13 | INTEHEAD [01.12.2007]  |
+14 | PRESSURE [...]         |
+15 | SWAT     [...]         |
+16 | OIL_DEN  [...]         |
+   |------------------------+
+12 | SEQNUM   [40]          |  E
+13 | INTEHEAD [01.12.2009]  |
+14 | PRESSURE [...]         |
+15 | SWAT     [...]         /
+
```
pgdr: Indiced of last view is off, should be 17,18,19,20.

python/python/ert/ecl/ecl_grav.py
```diff
@@ -67,45 +66,49 @@ def __init__( self, grid, init_file ):
...
+        to load the @restart_view argument is:
 
            import datetime
            import ert.ecl.ecl as ecl
            ...
            ...
            date = datetime.datetime( year , month , day )
-           restart_file1 = ecl.EclFile.restart_block( "ECLIPSE.UNRST" , dtime = date)
-           restart_file2 = ecl.EclFile.restart_block( "ECLIPSE.UNRST" , report_step = 67 )
+           restart_view1 = ecl.EclFile.restart_block( "ECLIPSE.UNRST" , dtime = date)
```
pgdr: `EclFile.restart_block` is deprecated and this should be updated to the new way of getting views.

pgdr:
`ecl_file.py:463` is `kw = self.__iget_named( kw_name , index )`.  The function `__iget_named` has been moved out of `ecl_file.py` and in to `ecl_file_view.py:26`.

libecl/src/ecl_grid.c
```diff
@@ -2656,32 +2656,28 @@ static void ecl_grid_init_nnc(ecl_grid_type * main_grid, ecl_file_type * ecl_fil
...
-        {
-          ecl_grid_type * grid = (lgr_nr > 0) ? ecl_grid_get_lgr_from_lgr_nr(main_grid, lgr_nr) : main_grid;
-          ecl_grid_init_nnc_cells(main_grid, grid , nncg, nncl);
-        }
+    if (ecl_file_view_has_kw(lgr_view , NNCL_KW)) {
+      const ecl_kw_type * nncl = ecl_file_view_iget_named_kw(lgr_view, NNCL_KW, 0);
+      const ecl_kw_type * nncg = ecl_file_view_iget_named_kw(lgr_view, NNCG_KW, 0);
+      {
+        ecl_grid_type * grid = (lgr_nr > 0) ? ecl_grid_get_lgr_from_lgr_nr(main_grid, lgr_nr) : main_grid;
+        ecl_grid_init_nnc_cells(main_grid, grid , nncg, nncl);
```
pgdr: Do you want to free `lgr_view`?

libecl/tests/ecl_nnc_test.c
```diff
@@ -36,19 +37,17 @@ void test_scan( const char * grid_filename) {
   for (block_nr = 0; block_nr < ecl_file_get_num_named_kw( grid_file , NNCHEAD_KW ); block_nr++) {
     ecl_grid_type * lgr = ecl_grid;
     int lgr_nr;
-    ecl_file_push_block( grid_file );
-    ecl_file_select_block(grid_file , NNCHEAD_KW , block_nr );
-    
+    ecl_file_view_type * nnc_view = ecl_file_alloc_global_blockview(grid_file, NNCHEAD_KW, block_nr);
```
pgdr: should this be freed too?

libecl/src/ecl_file_view.c
```diff
@@ -0,0 +1,785 @@
...
+int ecl_file_view_get_num_named_kw(const ecl_file_view_type * ecl_file_view , const char * kw) {
+  if (hash_has_key(ecl_file_view->kw_index , kw)) {
+    const int_vector_type * index_vector = hash_get(ecl_file_view->kw_index , kw);
+    return int_vector_size( index_vector );
+  } else
+    return 0;
+}
+
+void ecl_file_view_fwrite( const ecl_file_view_type * ecl_file_view , fortio_type * target , int offset) {
```
pgdr: Not `ecl_file_view_fwrite_fortio`?

libecl/include/ert/ecl/ecl_file.h
```diff
@@ -116,26 +101,35 @@ extern "C" {
...
+  ecl_file_view_type * ecl_file_get_restart_view( ecl_file_type * ecl_file , int input_index, int report_step , time_t sim_time, double sim_days);
+  ecl_file_view_type * ecl_file_get_summary_view( ecl_file_type * ecl_file , int report_step );
 
 /*****************************************************************/
 /* SUMMARY FILES */
 
-  bool             ecl_file_select_smryblock( ecl_file_type * ecl_file , int ministep_nr );
-  ecl_file_type  * ecl_file_open_smryblock( const char * filename , int ministep_nr , int flags);
+//  bool             ecl_file_select_smryblock( ecl_file_type * ecl_file , int ministep_nr );
+//  ecl_file_type  * ecl_file_open_smryblock( const char * filename , int ministep_nr , int flags);
```
pgdr: Have these lines been kept for a reason?

libecl/src/ecl_file_view.c
```diff
@@ -0,0 +1,785 @@
...
+    fprintf(stream , "%-8s %7d:%s\n",
+            ecl_file_kw_get_header( file_kw ) ,
+            ecl_file_kw_get_size( file_kw ) ,
+            ecl_util_get_type_name( ecl_file_kw_get_type( file_kw )));
+  }
+}
+
+
+ecl_file_view_type * ecl_file_view_alloc_blockview2(const ecl_file_view_type * ecl_file_view , const char * start_kw, const char * end_kw, int occurence) {
+  if ((start_kw == NULL) || ecl_file_view_get_num_named_kw( ecl_file_view , start_kw ) > occurence) {
```
pgdr: could you turn this into an early-return?

That is, if not that thing `(start_kw == NULL) || ecl_file_view_get_num_named_kw( ecl_file_view , start_kw ) > occurence`, return NULL?  Then everything else can be _unindented_

pgdr: Something like
```c
if ((start_kw != NULL) && ecl_file_view_get_num_named_kw( ecl_file_view , start_kw ) <= occurence)
  return NULL
```

libecl/include/ert/ecl/ecl_file_view.h
```diff
@@ -0,0 +1,74 @@
+#ifndef ERT_ECL_FILE_VIEW_H
```
pgdr: missing copyright headers

libecl/src/ecl_file_view.c
joakim-hove: Writing to fortio is the only possibility, so I thought it was not necessary to include it as part of the name.

libecl/src/ecl_file_view.c
```diff
@@ -0,0 +1,647 @@
...
+  }
+}
+
+
+int ecl_file_view_seqnum_index_from_sim_time( ecl_file_view_type * parent_map , time_t sim_time) {
+  int num_seqnum = ecl_file_view_get_num_named_kw( parent_map , SEQNUM_KW );
+  int seqnum_index = 0;
+  ecl_file_view_type * seqnum_map;
+
+  while (true) {
```
joakim-hove: I guess we could, but what is to gain by that? Personally I kind of like `while (true) { ... }` loops - then it is extremely obvious that the loop body must contain explicit break/return statements. In my opinion that is less clear with a for loop.

joakim-hove: I guess we could, but what is to gain by that? Personally I kind of like `while (true) { ... }` loops - then it is extremely obvious that the loop body must contain explicit break/return statements. In my opinion that is less clear with a for loop.

libecl/src/ecl_grid.c
joakim-hove: Yes

joakim-hove:
Thank you for a good review, will fix these comments as soon as I can.

libecl_well/tests/well_state_load_missing_RSEG.c
```diff
@@ -50,13 +51,13 @@ int main(int argc , char ** argv) {
     for (global_well_nr = 0; global_well_nr < header->nwells; global_well_nr++) {
       well_state_type * well_state = well_state_alloc(well_name , global_well_nr , open , type , report_nr , valid_from);
       test_assert_true( well_state_is_instance( well_state) );
-      well_state_add_connections( well_state , grid , rst_file , 0 );
-      well_state_add_MSW( well_state , rst_file , global_well_nr , load_segment_information );
+      well_state_add_connections2( well_state , grid , rst_view , 0 );
```
joakim-hove: Yes - but I think Resinsight is using the old `well_state_add_connections()` - so a rename was warranted :-(

joakim-hove:
OK - all the comments except: https://github.com/Ensembles/ert/pull/1250#pullrequestreview-3894509 and https://github.com/Ensembles/ert/pull/1250#pullrequestreview-3904116

libecl/src/ecl_file_view.c
pgdr: Well, I could agree, but in this example, the bottom of the while-loop is

```c
if (num_seqnum == seqnum_index)
      return -1;
```
and the `seqnum_index++` is hidden inside an `else` clause (that's always invoked except the last time).  Hence this is for all purposes a `for`-loop.

I'm not going to insist that you change anything, but I will insist that, as it is now, this is a manually driven for-loop.

pgdr:
That is good enough for me.  I will merge this when it greens.

python/python/ert/ecl/ecl_file.py
```diff
@@ -468,53 +389,7 @@ def iget_kw( self , index , copy = False):
...
-        if index < self.num_named_kw( kw_name ):
-            kw = self.__iget_named( kw_name , index )
-            if copy:
-                return EclKW.copy( kw )
-            else:
-                return kw
-        else:
-            raise KeyError('Asked for occurence:%d of %d of keyword "%s".'
-                           % (index  , self.num_named_kw( kw_name ), kw_name))
+        return self.global_view.iget_named_kw( kw_name , index , copy )
```
pgdr: `ecl_file_view` has no function `iget_named_kw`.  That function is protected (there is a function called `__iget_named_kw`), but it looks like you should have used `self.global_view[kw_name]` or something like that.

pgdr: Sorry, I meant `self.global_view[kw_name][index]`.  However, I don't know what to make of `copy`.

pgdr: But then you have to change `test_ecl_file_statoil.py:59` from `KeyError` to `IndexError`.

Alternatively, make `EclFile.iget_named_kw`  check for length of keyword and throw `KeyError` if `index >= n`.

joakim-hove:
OK - I'll merge this when it is green.

PR #1251, opened 2016-09-29
[WIP] Write and read EXEC_HOST from bjobs

pgdr:
* Call bjobs with a given jobid, output to tmp file
* parse said file and return stringlist of hostnames

### DIFF
--- a/libjob_queue/src/lsf_driver.c
+++ b/libjob_queue/src/lsf_driver.c
@@ -231,21 +231,29 @@ int lsf_job_parse_bsub_stdout(const char * bsub_cmd, const char * stdout_file) {
 /**
  * Assumes fname points to a file with content "hname1:hname2:hname3" as written by lsf_job_write_bjobs_to_file
  */
-stringlist_type * lsf_job_parse_bjobs_file(const char* fname) {
+stringlist_type * lsf_job_alloc_parse_hostnames(const char* fname) {
   FILE *stream = util_fopen(fname, "r");
-  stringlist_type * hosts = NULL;
+
   bool at_eof = false;
   while (!at_eof) {
     char * line = util_fscanf_alloc_line(stream, &at_eof);
     if (line != NULL) {
-      hosts = stringlist_alloc_from_split(line, ":"); // bjobs uses : as std. delimiter
+      stringlist_type * hosts = stringlist_alloc_from_split(line, ":"); // bjobs uses : as std. delimiter
+
+      for (int i = 0; i < stringlist_get_size(hosts); i++) {
+        const char * host = stringlist_iget(hosts, i);
+        stringlist_type * h = stringlist_alloc_from_split(host, "*");
+        stringlist_iset_copy(hosts, i, stringlist_iget(h, stringlist_get_size(h) - 1)); // hostname 4*be-lsf01 -> be-lsf01
+        stringlist_free(h);
+      }
+
       free(line);
       fclose(stream);
       return hosts;
     }
   }
   fclose(stream);
-  return hosts;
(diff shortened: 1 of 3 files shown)


libjob_queue/src/lsf_driver.c
```diff
@@ -228,6 +228,51 @@ int lsf_job_parse_bsub_stdout(const char * bsub_cmd, const char * stdout_file) {
   return jobid;
 }
 
+/**
+ * Assumes fname points to a file with content "hname1:hname2:hname3" as written by lsf_job_write_bjobs_to_file
+ */
+stringlist_type * lsf_job_parse_bjobs_file(const char* fname) {
```
joakim-hove: 1. This function can either return `NULL` or a `stringlist_type` instance; safer to return an empty list than `NULL`.
2. Functions which allocate memory which must be later free'd should have `alloc` as part of their name.
3. The purpose of this function is to extract hostnames, not really general bjobs parsing - I would suggest e.g. `lsf_job_parse_hostnames`.

libjob_queue/src/lsf_driver.c
```diff
@@ -228,6 +228,51 @@ int lsf_job_parse_bsub_stdout(const char * bsub_cmd, const char * stdout_file) {
...
+  FILE *stream = util_fopen(fname, "r");
+  stringlist_type * hosts = NULL;
+  bool at_eof = false;
+  while (!at_eof) {
+    char * line = util_fscanf_alloc_line(stream, &at_eof);
+    if (line != NULL) {
+      hosts = stringlist_alloc_from_split(line, ":"); // bjobs uses : as std. delimiter
```
joakim-hove: The sanitization of the `bjobs` output must also remove the trailing `n*` in front of the hostnames; i.e. for a parallell running with four CPU's on two hosts the raw output from `bjobs` can be:
```txt
4*host1:4*host2
```
we want to extract that to `{"host1", "host2"}`. Not sure if a "1*" is present for a scalar job.

libjob_queue/src/lsf_driver.c
```diff
@@ -228,6 +228,51 @@ int lsf_job_parse_bsub_stdout(const char * bsub_cmd, const char * stdout_file) {
...
+  const char * cmd = util_alloc_sprintf("%s %s -o '%s' %d", bjobs_cmd, noheader, fields, jobid);
+
+  const char * tmp_file = util_alloc_tmp_file("/tmp", "ert_job_exec_host", true);
+
+  if (driver->submit_method == LSF_SUBMIT_REMOTE_SHELL) {
+    char ** argv = util_calloc(2, sizeof *argv);
+    argv[0] = driver->remote_lsf_server;
+    argv[1] = cmd;
+    util_spawn_blocking(driver->rsh_cmd, 2, (const char **) argv, tmp_file, NULL);
+    free(argv[1]);
```
joakim-hove: Cleaner to call `free(cmd)` at the bottom of the function.

libjob_queue/src/lsf_driver.c
pgdr: Thank you.  I have addressed all these.  We now never return NULL and the function is called `lsf_job_alloc_parse_hostnames`.

libjob_queue/src/lsf_driver.c
pgdr: Thanks, added that feature together with tests.

libjob_queue/src/lsf_driver.c
pgdr: Fixed.  Thanks.

pgdr:
All comments addressed.  Thank you for the review.

PR #1256, opened 2016-10-04
Fixed LSF blacklist syntax

pgdr:
We now use syntax
```
bsub ... -R "... select[hname!='node1' && ... && hname!='noden'] ... " ...
```

### DIFF
--- a/libjob_queue/src/lsf_driver.c
+++ b/libjob_queue/src/lsf_driver.c
@@ -346,33 +346,37 @@ stringlist_type * lsf_driver_alloc_cmd(lsf_driver_type * driver ,
     quoting.
   */
 
-  stringlist_type * excludes = stringlist_alloc_new();
-  char * select = NULL;
-  if (driver->exclude_hosts != NULL) {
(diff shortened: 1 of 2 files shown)


libjob_queue/src/lsf_driver.c
```diff
@@ -346,13 +346,34 @@ stringlist_type * lsf_driver_alloc_cmd(lsf_driver_type * driver ,
     quoting.
   */
 
-  if  (driver->resource_request != NULL) {
-    if (driver->submit_method == LSF_SUBMIT_REMOTE_SHELL)
-      quoted_resource_request =util_alloc_sprintf("\"%s\"" , driver->resource_request);
-    else
-      quoted_resource_request = util_alloc_string_copy( driver->resource_request );
+  stringlist_type * excludes = stringlist_alloc_new();
```
joakim-hove: Could scope { ... } from lines 349 - 357 and call `stringlist_free( excludes )` at the bottom.

pgdr: Thanks, have pushed fix!

PR #1258, opened 2016-10-05
Blacklist failing LSF nodes

pgdr:
Added bool blacklist to `lsf_driver_kill_job`

### DIFF
--- a/libjob_queue/src/job_node.c
+++ b/libjob_queue/src/job_node.c
@@ -513,7 +513,7 @@ bool job_queue_node_update_status( job_queue_node_type * node , job_queue_status
         double runtime = job_queue_node_time_since_sim_start(node);
         if (runtime >= node->max_confirm_wait) {
           // max_confirm_wait has passed since sim_start without success; the job is dead
-          job_status_type new_status = JOB_QUEUE_EXIT;
+          job_status_type new_status = JOB_QUEUE_DO_KILL_NODE_FAILURE;
           status_change = job_queue_status_transition(status, current_status, new_status);
           job_queue_node_set_status(node, new_status);
         }
@@ -547,7 +547,6 @@ bool job_queue_node_status_transition( job_queue_node_type * node , job_queue_st
 void job_queue_node_set_max_confirmation_wait_time(job_queue_node_type * node, time_t time) {
   node->max_confirm_wait = time;
 }
-
 
 
 bool job_queue_node_status_confirmed_running(job_queue_node_type * node) {
--- a/libjob_queue/src/lsf_driver.c
+++ b/libjob_queue/src/lsf_driver.c
@@ -743,7 +743,7 @@ void lsf_driver_free_job(void * __job) {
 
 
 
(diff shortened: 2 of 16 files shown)


libjob_queue/src/job_node.c
```diff
@@ -549,6 +551,9 @@ void job_queue_node_set_max_confirmation_wait_time(job_queue_node_type * node, t
 }
 
 
+void job_queue_node_set_killed_blacklist(job_queue_node_type * node, bool blacklist) {
```
joakim-hove: Can not see a call to: `job_queue_node_set_killed_blacklist( node , true )`?

libjob_queue/src/lsf_driver.c
```diff
@@ -763,8 +763,25 @@ void lsf_driver_kill_job(void * __driver , void * __job) {
 
         free( argv[1] );
         free( argv );
-      } else if (driver->submit_method == LSF_SUBMIT_LOCAL_SHELL)
+      } else if (driver->submit_method == LSF_SUBMIT_LOCAL_SHELL) {
         util_spawn_blocking(driver->bkill_cmd, 1, (const char **) &job->lsf_jobnr_char, NULL, NULL);
+      }
+
+      if (blacklist) {
```
joakim-hove: If the blacklisting should be part of the kill the assembly of hostnames should be performed *before* the kill - otherwise I guess we could risk that the LSF server had started to tear down the hostname information.

joakim-hove:
OK; this is mainly fine, my only objection is that I feel uneasy about adding the `blacklist` flag to the `struct job_node`. I think an alternative (which I would prefer ...) would be to:

1. Add a new status `JOB_QUEUE_DO_KILL_NODE_FAILURE`.
2. In the main dispatch loop [here](https://github.com/Ensembles/ert/blob/master/libjob_queue/src/job_queue.c#L996) - add a entry for the new status `JOB_QUEUE_DO_KILL_NODE_FAILURE` which will *only* invoke the blacklisting[1] - and then change the status to `JOB_QUEUE_DO_KILL`.

That will be more in line with how the `job_queue` system works in general.

[1]: This will require a new driver callback - e.g. `node_failure` - which can be `NULL` for all drivers except the lsf driver.

pgdr:
@joakim-hove Thank you for your comments, I have tried to implemented them as I understood them.

Most importantly, I've added the flag
```c
    JOB_QUEUE_DO_KILL_NODE_FAILURE
```
which a job is assigned `if (runtime >= node->max_confirm_wait)`.

Jobs of this status should be handled by the following function (notice that we set it to `DO_KILL` after):

```c
static void job_queue_handle_DO_KILL_NODE_FAILURE(job_queue_type * queue, job_queue_node_type * node) {
  queue_driver_blacklist_node( queue->driver, node );
  job_queue_change_node_status(queue, node, JOB_QUEUE_DO_KILL);
}
```

where `queue_driver_blacklist_node` is implemented as
```c
  if (driver->driver_type == LSF_DRIVER)
    driver->blacklist_node(driver->data, job_data);
```

Round 2.  Fight!

PR #1265, opened 2016-10-12
Improve error messages related to keywords and filenames

pgdr:
This is a mild suggestion.

I find it difficulter to read `Does not have keyword:SWAT at ...` than to read `Does not have keyword "SWAT" at ...`.

The latter also has the advantage that it enables you to see number of spaces in the string.

### DIFF
--- a/python/python/ert/ecl/ecl_file.py
+++ b/python/python/ert/ecl/ecl_file.py
@@ -125,9 +125,9 @@ class EclFile(BaseCClass):
 
         if not OK:
             if dtime:
-                raise ValueError("Could not locate date:%02d/%02d/%4d in restart file: %s." % (dtime.day , dtime.month , dtime.year , filename))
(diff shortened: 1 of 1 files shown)


joakim-hove:
>I find it difficulter to read Does not have keyword:SWAT at ... than to read Does not have keyword "SWAT" at ....

Those changes are fine by me. In addition it seems you have done a `" -> '`  change. Personally I have a weak preference for `"` - since that is the quoter used in C/C++, but I have no strong feelings. Was there a specific reason for the quote change?

pgdr:
No, the PEP8 recommends avoiding escaping quotes in strings `\"` by choosing the other quote as the outer quote symbol.  I wanted to have `"filename"`, and thus needed to use `'` for the main string.

In addition I have a weak preference for `'`, but that is purely aesthetic (since the Python REPL prefers `'`).

PEP8 handles them equally (with the exception for docstrings, which should be `"""`).

PR #1271, opened 2016-10-13
Shuffled around with enkf_main_UPDATE

pgdr:
While debugging, I reshuffled some code in `enkf_main_UPDATE`.

There is _no change of semantics_ — I simply added two early returns, and extracted the rest to a separate function.

Do ask questions and provide feedback.  The sole purpose of this PR is to make the `enkf_main_UPDATE` function easier to read; if this PR does not do that, we'll close it and forget about it.

### DIFF
--- a/libenkf/src/enkf_main.c
+++ b/libenkf/src/enkf_main.c
@@ -1150,7 +1150,7 @@ static FILE * enkf_main_log_step_list(enkf_main_type * enkf_main, const int_vect
 /**
  * This is THE ENKF update function.  It should only be called from enkf_main_UPDATE.
  */
-void enkf_main_update__(enkf_main_type * enkf_main, const int_vector_type * step_list, enkf_fs_type * source_fs,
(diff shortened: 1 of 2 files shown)


libenkf/src/enkf_main.c
```diff
@@ -1131,171 +1131,194 @@ static void enkf_main_analysis_update( enkf_main_type * enkf_main ,
...
+  free(log_file);
+  return log_stream;
+}
 
 
-bool enkf_main_UPDATE(enkf_main_type * enkf_main , const int_vector_type * step_list, enkf_fs_type * source_fs , enkf_fs_type * target_fs , int target_step , run_mode_type run_mode) {
+/**
+ * This is THE ENKF update function.  It should only be called from enkf_main_UPDATE.
+ */
+void enkf_main_update__(enkf_main_type * enkf_main, const int_vector_type * step_list, enkf_fs_type * source_fs,
```
joakim-hove: I guess this should be `static`?

joakim-hove:
OK - will merge when `enkf_main_update__( )` has been marked as static.

PR #1272, opened 2016-10-17
added: allow building the c++ bindings using std=c++14

akva2:
relies on cmake >= 3.1. for older, it falls back to the old
behavior of hardcoding c++11


akva2:
hmm a lot of tests failed on osx but i suspect that is expected since it seems to allow failures there. please consider if there are more failures there or nto.

PR #1274, opened 2016-10-18
LSF checks for nonempty exclude_hosts and stringlist join returns ""

### LINKED ISSUE #1273: Failed when submitting to LSF
Error message:

```
Error in select section: Error near "null" : incorrect section name, resource name, or resource function name. Job not submitted.

** ERROR ** Failed when submitting to LSF - will try again.
```

pgdr:
This PR contains two changes:

* LSF driver checks that we have excluded hosts before applying `select[%s]` flag
* `stringlist_alloc_joined_string` (and `...substring`) now returns `""` on an empty list, and never `NULL`

This should fix #1273

### DIFF
--- a/libjob_queue/src/lsf_driver.c
+++ b/libjob_queue/src/lsf_driver.c
@@ -357,23 +357,26 @@ stringlist_type * lsf_driver_alloc_cmd(lsf_driver_type * driver ,
 
     char * excludes_string = NULL;
     char * req = "";
+    char * resreq = NULL;
 
-    if (driver->resource_request != NULL) {
-      char * resreq = util_alloc_string_copy(driver->resource_request);
-      if (stringlist_get_size(select_list) > 0) {
+    if (stringlist_get_size(select_list) > 0) {
+      excludes_string = stringlist_alloc_joined_string(select_list, " && ");
+      if (driver->resource_request != NULL) {
+        resreq = util_alloc_string_copy(driver->resource_request);
         util_string_tr(resreq, ']', ' '); // remove "]" from "select[A && B]
         excludes_string = stringlist_alloc_joined_string(select_list, " && ");
         req = util_alloc_sprintf("%s && %s]", resreq, excludes_string);
       } else {
-        req = util_alloc_sprintf("%s", resreq);
-      }
-      util_free(resreq);
-    } else {
-      if (stringlist_get_size(select_list) > 0) {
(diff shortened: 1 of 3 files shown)


pgdr:
The main problem turned out to be two `select`-strings:

```
bsub ... -R select[hname!='node'] select[cs && Linux]
```

I had to expand the `driver->resource_request` with the exclude hosts list.

libjob_queue/src/lsf_driver.c
```diff
@@ -347,35 +347,43 @@ stringlist_type * lsf_driver_alloc_cmd(lsf_driver_type * driver ,
...
-    stringlist_type * qrr = stringlist_alloc_new();
-    if (select) {
-      stringlist_append_copy(qrr, select);
-      util_free(select);
-    }
-
-    if (driver->resource_request != NULL)
-      stringlist_append_copy(qrr, driver->resource_request);
+    char * excludes_string = NULL;
+    char * req = "";
```
pgdr: There are four cases here.  `driver->resource_request` can be empty/NULL, but is typically `select[cs && x86_64Linux]`, and `driver->exclude_hosts`  can be empty, or a list of hostnames (e.g. `[node1, node2]`).

In the end we are interested in a string `"select[cs && whatever && hname!=node1 && hname!=node2]"`.

What is done here, is to remove the `']'` from `driver->resource_request` and concatenate `stringlist_alloc_joined_string(select_list, " && ")` on the end of it, if it exists.

Or to create `"select[hname!=node1 && hname!=node2]"` if no resource_request.

pgdr:
(ps, I would love a better way of doing this!)

pgdr:
Thank you for the review.  Comments addressed.

libjob_queue/src/lsf_driver.c
```diff
@@ -347,35 +347,46 @@ stringlist_type * lsf_driver_alloc_cmd(lsf_driver_type * driver ,
...
       }
-      const char * excludes_string = stringlist_alloc_joined_string(excludes, " && ");
-      select = util_alloc_sprintf("select[%s]", excludes_string);
     }
 
-    stringlist_type * qrr = stringlist_alloc_new();
-    if (select) {
-      stringlist_append_copy(qrr, select);
-      util_free(select);
+    char * excludes_string = NULL;
```
joakim-hove: This could be inside the `if (stringlist_get_size( select_list ) > 0) {` block.

PR #1277, opened 2016-10-20
Doxygen

pgdr:
This PR adds a doxygen cfg file, as well as creates a target with CMake `make doxy`.


docs/doxygen.cfg
```diff
@@ -0,0 +1,42 @@
+# This is a slightly modified standard doxygen config file
+
+DOXYFILE_ENCODING      = UTF-8
+PROJECT_NAME           = "Ert"
+PROJECT_NUMBER         = 2.1
```
joakim-hove: This should be created with the cmake `configure_file( )` functionality, getting the version numbers from the cmake variables [here](https://github.com/Ensembles/ert/blob/master/CMakeLists.txt#L10).

joakim-hove:
Good change - will merge as soon as the version number is handled automagically.

pgdr:
You are right, it's now writing in-source.  I added the output folder to the `.gitignore`, but if you think it's better to output to `build`-directory, then I'll change it to that.

joakim-hove:
I would really prefer out og tree building.

pgdr:
Comments addressed:  build out-of-tree and use `${ERT_VERSION_MAJOR}.${ERT_VERSION_MINOR}` for version.

PR #1280, opened 2016-10-25
Making sure the ertshell is creating the run path

agchitu:
The ertshell has a different entry for running the simulations than the
gert and others. Added code to create the run path and run the possibly
defined pre-simulation workflow.

The current implementation is not yet 100% neat though since it is
hardcoding the iteration number to 0 (zero). Need help! If I set the
number of iteration_count the createRunPath is creating only the last
subfolder namely iter_4 which ends up with a nice error.

### DIFF
--- a/python/python/ert_gui/shell/simulations.py
+++ b/python/python/ert_gui/shell/simulations.py
@@ -35,13 +35,10 @@ class Simulations(ErtShellCollection):
         now = time.time()
         print("Ensemble Experiment started at: %s" % datetime.now().isoformat(sep=" "))
 
-        print("Create the run path!")
-        iteration_count = self.ert().analysisConfig().getAnalysisIterConfig().getNumIterations()
-        active_realization_mask = BoolVector.createActiveMask("0-%d" % self.ert().getEnsembleSize())
-        # Note that 0(zero) is used and not iteration_count
-        simulation_runner.createRunPath(active_realization_mask, 0)
+        iteration_count = 0
+        active_realization_mask = BoolVector(default_value = True, initial_size = self.ert().getEnsembleSize())
+        simulation_runner.createRunPath(active_realization_mask, iteration_count)
 
-        print("Ensemble Experiment pre processing!")
         simulation_runner.runWorkflows(HookRuntime.PRE_SIMULATION)
 
         print("Start simulations!")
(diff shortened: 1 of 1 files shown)


joakim-hove:
Beautiful PR!

python/python/ert_gui/shell/simulations.py
```diff
@@ -33,9 +33,18 @@ def ensembleExperiment(self, line):
         simulation_runner = EnkfSimulationRunner(self.ert())
 
         now = time.time()
-
         print("Ensemble Experiment started at: %s" % datetime.now().isoformat(sep=" "))
 
+        print("Create the run path!")
+        iteration_count = self.ert().analysisConfig().getAnalysisIterConfig().getNumIterations()
+        active_realization_mask = BoolVector.createActiveMask("0-%d" % self.ert().getEnsembleSize())
```
joakim-hove: This should be `"0-%d" % self.ert().getEnsembleSize() - 1` - but even better:
```Python
active_realization_mask = BoolVector( default_value = True , initial_size = self.ert().getEnsembleSize())
```
not any reason to go through the string based construction.

python/python/ert_gui/shell/simulations.py
```diff
@@ -33,9 +33,18 @@ def ensembleExperiment(self, line):
         simulation_runner = EnkfSimulationRunner(self.ert())
 
         now = time.time()
-
         print("Ensemble Experiment started at: %s" % datetime.now().isoformat(sep=" "))
 
+        print("Create the run path!")
```
joakim-hove: Remove the `print( ... )` statements - only the final `print("Start simulations")`should be retained.

python/python/ert_gui/shell/simulations.py
```diff
@@ -33,9 +33,18 @@ def ensembleExperiment(self, line):
...
+        iteration_count = self.ert().analysisConfig().getAnalysisIterConfig().getNumIterations()
```
agchitu: I'll do this as well, but I will keep the comment about the iteration_count so that the next person that looks as it will know that something is not yet watertight.

python/python/ert_gui/shell/simulations.py
agchitu: Sure.

joakim-hove:
>The current implementation is not yet 100% neat though since it is hardcoding the iteration number to 0 (zero). 

That is OK - we do not have any context to give us a better value than zero here. My suggestion is that you just do:
```Python
iter = 0
simulation_runner.createRunPath(active_realization_mask, iter)
```
then it is very clear to everyone that the iter value has been hardcoded, and we can at some later stage decide to take it as an input argument.

joakim-hove:
OK - I'm merging; thank you!

PR #1288, opened 2016-10-30
Removed queue system setting dependencies.

joakim-hove:
Have removed the requirement to set options directly on the drivers,
i.e. LSF_MAX_RUNNING and such. The general interface going through
QUEUE_OPTION should be used:

   QUEUE_OPTION LSF MAX_RUNNING 88


jokva:
Pure deletion! Nice.

PR #1291, opened 2016-11-01
Gen data local

joakim-hove:


joakim-hove:
This has been tested by end users- merging.

PR #1292, opened 2016-11-02
More tests for Geertsma kernel

kjellkongsvik:
2 source points, 2 vintages

### DIFF
--- a/python/tests/core/ecl/test_geertsma.py
+++ b/python/tests/core/ecl/test_geertsma.py
@@ -18,12 +18,13 @@ def create_init(grid, case):
         porv.fwrite(f)
 
 
-def create_restart(grid, case):
+def create_restart(grid, case, p1, p2=None):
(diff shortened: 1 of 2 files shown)


python/tests/core/ecl/test_geertsma.py
```diff
@@ -61,3 +72,37 @@ def test_geertsma_kernel():
...
+
+            subsidence = EclSubsidence(grid, init)
+            subsidence.add_survey_PRESSURE("S1", restart_view1)
+            subsidence.add_survey_PRESSURE("S2", restart_view2)
+
+            youngs_modulus = 5E8
+            poisson_ratio = 0.3
+
+            dz1 = subsidence.evalGeertsma("S1", None, (1000, 1000, 0), youngs_modulus, poisson_ratio)
+            np.testing.assert_almost_equal(dz1, -5.538064265738908e-05, 15)
```
joakim-hove: Reduce the precision to e.g. 15 decimals; with this extreme precision we might get false positives due to e.g. platform differences.

PR #1294, opened 2016-11-03
Make doc

joakim-hove:
Changes in documentation configuration/generation.


jokva:
Slightly confused about the symlink generation into the binary dir. What's the rationale?

joakim-hove:
> Slightly confused about the symlink generation into the binary dir. What's the rationale?

The point is that source directory as seen from the sphinx tool consists of three parts:

1. Normal rst files with handwritten documentation from the repository.
2. Auto generated api documentation.¨
3. Handwritten rst files from an external source.

Of these three only the first is source code in the sense of the current repository, we therefor create a directory `doc-src/` in the build directory and populate that with symlinks and generated api documentation, and then present that directory to the sphinx documentation builder as a source code directory,

jokva:
Oh right. So to avoid copying really? All good then.

PR #1296, opened 2016-11-06
Smspec restart

joakim-hove:
Ready to be merged.


pgdr:
I think the PR message has potential for improvement.  What is this PR solving?  What has warranted it?  (Is there a task for it?)

joakim-hove:
Thank you for review. There is no task. Requested by users.

The purpose is the following: When a eclipse case is restarted information about the initial case is embedded in the summary result files. When loading summary we will also load the initial case, previously the initial and current case had to have exactly the same header layout, with this PR there is a mapping between indices.

PR #1297, opened 2016-11-08
Resurrected block_fs utilities from the past.

joakim-hove:


jokva:
By doing this the history for block_fs is "lost" i.e. reading the log *for the file itself* won't take you back further than this. Is this acceptable? If yes, merge ahoy. If not, it can probably be restored (i.e. its deletion commit reverted) and then moved to the new directory (without devel/).

PR #1298, opened 2016-11-09
Added Copying file with license

blattms:
The current situation is the following:
Some source files contain copyright information, others do not (e.g. cmake/*, lib*/*, and python/python/*).
To get a full list of those file, use `licensecheck -r .|sed "s/\([^:]*\): \([^:]*\)/\2 : \1/"| sort -n| grep -i unknown`.

This PR at least adds a COPYING file with the GPL v3+ licence. Otherwise people need to browse the homepage to find out.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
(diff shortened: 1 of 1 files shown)


joakim-hove:
Thank you.

blattms:
Wow, the man who merges faster than his shadow. Thanks!

PR #1299, opened 2016-11-09
Added manpage for ecl_summary as requested by Debian policy.

blattms:
This is only a rudimentary manpage from a template produced by help2man.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -218,6 +218,12 @@ if (BUILD_PYTHON)
       if(ERT_DOC)
          add_subdirectory( docs )
       endif()
+
+      # Always install the man page.
+      include(GNUInstallDirs)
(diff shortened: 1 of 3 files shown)


joakim-hove:
Very good - thank you. Should the man page be installed also as part of the RedHat packaging?

blattms:
Yes it should. I forgot to cherry-pick that commit, sorr. Should be
added now.

joakim-hove:
> Should be added now.

I was expecting there should be a change in https://github.com/Ensembles/ert/blob/master/redhat/ert.ecl.spec - but maybe not?

blattms:
Did not think about the packages. Will try to add there, too.

blattms:
I do not feel qualified enough to patch the redhat stuff.
Maybe somebody else can tell it that usr/share/man/man1/ecl_summary.1
needs to be part of the package?

joakim-hove:
Fine - merging.

PR #1300, opened 2016-11-09
Allow for user provided CMAKE_C_FLAGS on linux

blattms:
This was not possible before and is e.g. needed to use flags recommended by packaging systems of Linux distributions (read: Debian).


pgdr:
Thanks.

PR #1302, opened 2016-11-09
Fix diagnostics

bska:
This change-set fixes a few diagnostic messages that I got following PR #1300.  It is mostly adding explicit `void` keywords to declarations of functions that do not take arguments.

Commit 5d6eb7f is potentially contentious.  The commit assumes that the macro in question (i.e., `UTIL_IS_INSTANCE_FUNCTION`) will only ever be processed by a C compiler and that we can rely on `void*` being implicitly convertible to all other (object) pointer types.  If the macro is supposed to be usable in a C++ context, then the body of the macro must be different.

### DIFF
--- a/libecl/include/ert/ecl/ecl_coarse_cell.h
+++ b/libecl/include/ert/ecl/ecl_coarse_cell.h
@@ -27,7 +27,7 @@ extern "C" {
   typedef struct ecl_coarse_cell_struct    ecl_coarse_cell_type;
 
   bool                   ecl_coarse_cell_equal( const ecl_coarse_cell_type * coarse_cell1 , const ecl_coarse_cell_type * coarse_cell2);
-  ecl_coarse_cell_type * ecl_coarse_cell_alloc( );
+  ecl_coarse_cell_type * ecl_coarse_cell_alloc( void );
   void                   ecl_coarse_cell_update( ecl_coarse_cell_type * coarse_cell , int i , int j , int k , int global_index );
   void                   ecl_coarse_cell_free( ecl_coarse_cell_type * coarse_cell );
   void                   ecl_coarse_cell_free__( void * arg );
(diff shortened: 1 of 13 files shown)


joakim-hove:
Thank you - will merge. Just a quick question first: 

> The commit assumes that the macro in question (i.e., UTIL_IS_INSTANCE_FUNCTION) will only ever be processed by a C compiler and that we can rely on void* being implicitly convertible to all other (object) pointer types. If the macro is supposed to be usable in a C++ context, then the body of the macro must be different.

I guess the `UTIL_IS_INSTANCE_FUNCTION` will only be processed by a C compiler; but can out of the top of your head suggest changes required to make i C++ safe?

bska:
> I guess the `UTIL_IS_INSTANCE_FUNCTION` will only be processed by a C compiler; but can out of the top of your head suggest changes required to make i C++ safe?

Sure, but I first need to clarify my wording.  The **macro** itself is safe in either language, but it is the resulting function (i.e., the `<T>_is_instance()` function that results from substituting macro arguments) that assumes a C compiler.  In a C++ context, the body of `<T>_is_instance()` needs to use the explicit conversion
```C++
(const type ## _type *) __arg
```
In other words, we'd need the change
```diff
diff --git a/libert_util/include/ert/util/type_macros.h b/libert_util/include/ert/util/type_macros.h
index d1075d9..ec77818 100644
--- a/libert_util/include/ert/util/type_macros.h
+++ b/libert_util/include/ert/util/type_macros.h
@@ -49,7 +49,7 @@ bool type ## _is_instance( const void * __arg ) {          \
    if (__arg == NULL)                                      \
       return false;                                        \
    else {                                                  \
-      const type ## _type * arg =                   __arg; \
+      const type ## _type * arg = (const type ## _type *) __arg; \
       if ( arg->__type_id == TYPE_ID)                      \
          return true;                                      \
       else                                                 \
```
relative to 5d6eb7f.

By the way the symbol `__arg` is, strictly speaking, reserved for the implementation.  User-level identifiers are not allowed to have two consecutive underscores anywhere.  It is unlikely that this will ever become a problem.

joakim-hove:
> In other words, we'd need the change

OK - thanks; I'll do that!

PR #1308, opened 2016-11-14
Added geo_surface_alloc_new in libgeometry

pgdr:
* added function `geo_surface_alloc_new` allowing creating empty surface
* added corresponding mapping in `surface.py`
* added tests for creating new surfaces

_(Internal ERT-1107)_

### DIFF
--- a/libgeometry/src/geo_surface.c
+++ b/libgeometry/src/geo_surface.c
@@ -196,6 +196,30 @@ void geo_surface_fprintf_irap_external_zcoord( const geo_surface_type * surface,
   geo_surface_fprintf_irap__( surface , filename , zcoord );
 }
 
+geo_surface_type  * geo_surface_alloc_new( int nx,        int ny,
+                                           double xinc,   double yinc,
+                                           double xstart, double ystart,
+                                           double angle ) {
+    geo_surface_type * surface = geo_surface_alloc_empty( true );
+
+    surface->origo[0]  = xstart;
+    surface->origo[1]  = ystart;
+    surface->rot_angle = angle * __PI / 180.0;
+    surface->nx = nx;
+    surface->ny = ny;
+
+    surface->vec1[0] = xinc * cos( surface->rot_angle ) ;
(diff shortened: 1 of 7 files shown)


pgdr:
(Added commit for whitespaces & spelling changes, so review one commit at a time.)

libgeometry/src/geo_surface.c
```diff
@@ -196,6 +196,29 @@ void geo_surface_fprintf_irap_external_zcoord( const geo_surface_type * surface,
   geo_surface_fprintf_irap__( surface , filename , zcoord );
 }
 
+geo_surface_type  * geo_surface_alloc_new( int ny, double xinc, double yinc,
```
joakim-hove: The order of arguments here was quite funny:

1. `x` is *always* before `y`.
2. Either all `x` arguments and then all `y` arguments as: `nx, xstart, xinc, ny, ystart,yinc` or alternating: `nx,ny,xstart,ystart,xinc,yinc`

python/python/ert/geo/surface.py
```diff
@@ -44,18 +45,27 @@ class Surface(BaseCClass):
...
+                                      nx=None, angle=None, internal_z=None):
         """
         This will load a irap surface from file. The surface should
         consist of a header and a set z values.
         """
-        if os.path.isfile( filename ):
-            c_ptr = self._alloc(filename , True)
-            super(Surface , self).__init__(c_ptr)
+        if filename is not None:
+            filename = str(filename)
```
joakim-hove: The `internal_z` should in 99/100 cases be `True`. At least it should default to `True` - alternatively it could even be hardcoded as `internal_z = True` inside the method.

libgeometry/src/geo_surface.c
joakim-hove: OK - I see; well - nothing to do about the file format, but I still think we should make the method signature sane.

thorvaldj:
Nice and tidy commits.  Ready for merge :-)

pgdr:
Thanks for the review.

PR #1314, opened 2016-11-16
Run Analysis tool.

jokva:
Adds the Run Analysis tool to the main toolbar, which gives a quick way
of running analysis on the current case, available from the main gui.

Reuses the analysis selector used in the simulation configuration
panel.

1-full-ert
2-run-analysis
3-target-dialogue
4-analysis-selector


joakim-hove:
This is good - and to the specification; but oh my - the specification has changed! We are trying to reduce "state" in general, and the *current case* is one particular obnoxious example of state. So would be good if the **Run Analysis** got an extra input field with the name of the source case - this could naturally be initialized with the current case.

python/python/ert_gui/tools/run_analysis/run_analysis_tool.py
```diff
@@ -0,0 +1,67 @@
...
+from ert_gui.ertwidgets import resourceIcon
+from ert_gui.ertwidgets.closabledialog import ClosableDialog
+from ert_gui.tools import Tool
+from ert_gui.tools.run_analysis import RunAnalysisPanel
+
+
+class RunAnalysisTool(Tool):
+    def __init__(self):
+        super(RunAnalysisTool, self).__init__("Run Analysis", "tools/run_analysis", resourceIcon("ide/table_import"))
+        self.__run_widget = None
```
jepebe: Double underscore is out of fashion in ERT. :)

jepebe:
To follow up on Joakim's request the:

`ert_gui.ertwidgets.caseselector.CaseSelector`

Will do this. Just set `update_ert=False`

jokva:
Adress'd.

PR #1318, opened 2016-11-17
RML module log and run analysis

jokva:
Two small commits that smooths some RML related bumps.

* RML now logs by default (to rml_enkf.out)
* The RML module shows up in the **Run Analays** tool.


joakim-hove:
Ready for merge

PR #1321, opened 2016-11-21
Removed print Emitting ertChanged

pgdr:


pgdr:
Do we have access to a debug flag, so we can do `if DEBUG: print(...)`?

PR #1322, opened 2016-11-21
Added support for top of active cells

pgdr:
This PR adds a feature to `ecl_grid.py`:

* added function `top_active` that returns top of the active part of the reservoir in a given column, or raises if column is inactive


python/python/ert/ecl/ecl_grid.py
```diff
@@ -713,11 +714,29 @@ def depth( self , active_index = None , global_index = None , ijk = None):
...
+    def top_active( self, i, j ):
+        """
+        Top of the active part of the reservoir; in the column (@i , @j).
+        Returns nan if (i,j) column is inactive.
+        """
+        for k in range(self.getNZ()):
+            a_idx = self.get_active_index(ijk=(i,j,k))
+            if a_idx >= 0:
+                return self.top(active_index=a_idx)
+        return float('nan')
```
joakim-hove: I would prefer if the `top_active()` method called `self._get_top1A( )` directly instead of forwarding to `top( )`. As I see it the forwarding makes `top( )` considerably more cluttered?

PR #1324, opened 2016-11-22
OpenMP implementation of fwd_step

arielalmendral:


libanalysis/src/fwd_step_enkf.c
```diff
@@ -59,6 +62,7 @@ struct fwd_step_enkf_data_struct {
   long                       option_flags;
   double                     r2_limit;
   bool                       verbose;
+  int                        num_threads;
```
joakim-hove: Would it be possible to initialize this sensibly from e.g `/proc/cpuinfo`

arielalmendral: Detecting automatically the number of threads. The functionality can be overridden with the flag NUM_THREADS as before. 

Note the cmake changes. 

1. Now OpenMP is default on, which means that it will be in use for the eclipse grid slice functionality. In ecl_grid.c, function **ecl_grid_init_GRDECL_data** the number of threads is not set, so the default is 1?

2. find_package(OpenMP) is not working properly. The c++ compiler has indeed openmp, but this cmake module is saying that it is not there, and unsetting the OpenMP_CXX_FLAGS, necessary to link with openmp. Therefore this workaround that does not use find_package(OpenMP) and sets the flags directly.

CMakeLists.txt
```diff
@@ -82,20 +85,22 @@ if (MSVC)
...
 if (ERT_USE_OPENMP)
-   find_package(OpenMP)
-   if (OPENMP_FOUND) 
-       message(STATUS "Enabling OpenMP support")
-       # The actual use of OpenMP is only in the libecl library - the compile flags is only applied there.
-   else()
-       set( ERT_USE_OPENMP OFF )
-       message(STATUS "OpenMP package not found - OpenMP disabled")
-   endif()
+  find_package(OpenMP)
```
arielalmendral: Checking the presence of OpenMP AFTER find_package(CXX11Features) fails due to the CXX flags that are set by  find_package(CXX11Features).

PR #1325, opened 2016-11-23
Remove usage of CWrapper in all ert/ecl

pgdr:
commits:
1. minor whitespaces, can be ignored
1. deleted unused `geo_polygon` and rewrote `ecl_restart_file` to use `BaseCClass` over `CWrapper`
1. rewrote `fault_block` to use `BaseCClass` over `CWrapper`
1. rewrote `fault_block_layer` ...
1. rewrote `fault_block_collection` ...
1. rewrote `layer` ...

_(internal: ERT:1117→ERT-1108)_


python/python/ert/ecl/faults/fault_block.py
```diff
@@ -32,23 +31,37 @@ def __init__(self , i,j,k ,x,y,z):
...
 
-        
+
     def __str__(self):
         return "(%d,%d)" % (self.i , self.j)
 
 
 
 class FaultBlock(BaseCClass):
+    TYPE = "fault_block"
```
jepebe: TYPE_NAME

PR #1331, opened 2016-11-25
implemented full __getitem__ for GenData

pgdr:
Implemented `__getitem__` for data type `GenData` and modernized `GenKw`, `GenKwConfig`, and `GenDataFileType` (now using `EnkfPrototype`).

Implemented `__repr__` for the aforementioned classes.

### DIFF
--- a/python/python/ert/enkf/data/gen_data.py
+++ b/python/python/ert/enkf/data/gen_data.py
@@ -1,21 +1,31 @@
 from cwrap import BaseCClass, CWrapper
 from ert.util import DoubleVector
-from ert.enkf import ENKF_LIB
-
+from ert.enkf import EnkfPrototype
 
 
 class GenData(BaseCClass):
+    TYPE_NAME = "gen_data"
+    _alloc       = EnkfPrototype("void*  gen_data_alloc()", bind = False)
+    _free        = EnkfPrototype("void   gen_data_free(gen_data)")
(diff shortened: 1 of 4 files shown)


python/python/ert/enkf/data/gen_data.py
```diff
@@ -30,6 +30,18 @@ def getData(self):
         GenData.cNamespace().export_data( self , data )
         return data
 
+    def __getitem__( self, idx ):
+        ls = len(self)
+        if isinstance( idx, int ):
+            if idx < 0:
+                idx += ls
+            if 0 <= idx < ls:
+                return self.getData()[idx]
```
joakim-hove: The `GenData` instances can potentially be *large* - they are typically used for seismic data. That means that creating the  full data vector and then only extracting one element from that vector looks very costly. Should rather use `gen_data_iget_double( )` to get one scalar value.

python/python/ert/enkf/data/gen_data.py
```diff
@@ -30,6 +30,18 @@ def getData(self):
...
+            raise IndexError('List index out of range.')
+        if isinstance( idx, slice ):
+            vec = self.getData()
+            return [vec[i] for i in range(*idx.indices(ls))]
```
joakim-hove: With this implementation the slice return value will be a standard Python list; i.e. not the same type as self. I think that is the only possible way to implement slices for `GenData` - but I guess a comment could be added.

PR #1336, opened 2016-12-01
Pass max_runtime to the jobs.py file.

joakim-hove:
Uncertain if this should be merged? Let it linger in the first place.


pgdr:
Merge at will.

PR #1341, opened 2016-12-01
EnsemblePlot and PcaPlot is cwrapper-free

pgdr:


pgdr:
I think this is a rather uncontroversial PR which can be merged now that tests are green.

jepebe:
Are all of these classes still in use?

pgdr:
I don't know, but I think this can be merged, and then we can check whether they are used afterwards.  It's a bit hard to tell what's used and what's not, since this is exposed in our api.

Btw, I don't know why it says "some checks haven't completed yet", because if you go to Travis, it says:
```
git fetch origin +refs/pull/1341/merge
...
...
100% tests passed, 0 tests failed out of 304
Label Time Summary:
Python         = 565.38 sec
StatoilData    =   0.00 sec
Total Test time (real) = 748.96 sec
```

PR #1342, opened 2016-12-05
Default use OPENMP to off

joakim-hove:


arielalmendral:
Wondering, why did you have to turn off OpenMP?

joakim-hove:
> Wondering, why did you have to turn off OpenMP?

I'll call you.

PR #1344, opened 2016-12-05
Added test for Geir Evensens subsspace PR

pgdr:
This PR is a minor continuation from the work done by @geirev.

It implements two simple tests, one for verifying that we respect `USE_GE` and that the `initX` outputs different values.

Added a missing include `from ert.util import Matrix` in the original work.

### DIFF
--- a/libert_util/src/matrix.c
+++ b/libert_util/src/matrix.c
@@ -277,6 +277,16 @@ matrix_type * matrix_safe_alloc(int rows, int columns) {
   return matrix_alloc__( rows , columns , true );
 }
 
+matrix_type * matrix_alloc_identity(int dim) {
+  if (dim < 1)
+    util_abort("%s: identity matrix must have positive size. \n",__func__);
+
+  matrix_type * idty = matrix_alloc(dim, dim);
+  for (int i = 0; i < dim; ++i)
+    matrix_iset(idty, i, i, 1);
+  return idty;
+}
+
 /*****************************************************************/
 
 /**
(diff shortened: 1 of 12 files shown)


python/tests/ert/analysis/test_analysis_module.py
```diff
@@ -73,3 +74,87 @@ def test_create_internal(self):
...
+            for j in range(n):
+                m[(i,j)] = vals[idx]
+                idx += 1
+        return m
+
+    def _n_identity_mcs(self, n=6,s=3):
+        """return n copies of the identity matrix on s*s elts"""
+        return tuple([self._identity_mx(s) for i in range(n)])
+
+    def _identity_mx(self,s=3):
```
joakim-hove: Instead of this test special casing I would like to get a first class method to create a diagonal matrix; i.e. `matrix_alloc_identity( int size )` in C and a corresponding classmethod in Python.

libert_util/src/matrix.c
```diff
@@ -277,6 +277,20 @@ matrix_type * matrix_safe_alloc(int rows, int columns) {
   return matrix_alloc__( rows , columns , true );
 }
 
+matrix_type * matrix_alloc_identity(int dim) {
+  if (dim < 1)
+    util_abort("%s: identity matrix must have positive size. \n",__func__);
+
+  matrix_type * idty = matrix_alloc(dim, dim);
+  for (int i = 0; i < dim; ++i)
```
joakim-hove: The matrix is initialized to zero, so this can be simplified to:
```C
for (int i=0; i < dim; i++)
    matrix_iset( idty , i , i , 1 );
```

joakim-hove: And then - merge!

PR #1347, opened 2016-12-05
Some python3 initial work, and cwrap vs Python 3

pgdr:
1. python3 fixes
** fix `except Exception, err`
** remove use of spaceship operator `<>`
** make import test python 3 compatible; do not import `__pycache__`
1. renamed `registerType` to `_registerType`, do not use outside `cwrap` module!
1. removed tests from `cwrap` of functionality covered in `test_metacwrap`

### DIFF
--- a/python/python/cwrap/prototype.py
+++ b/python/python/cwrap/prototype.py
@@ -31,7 +31,7 @@ REGISTERED_TYPES = {}
 """:type: dict[str,TypeDefinition]"""
 
 
-def registerType(type_name, type_class_or_function, is_return_type=True, storage_type=None):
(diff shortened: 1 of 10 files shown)


python/python/cwrap/prototype.py
```diff
@@ -31,36 +31,36 @@ def __init__(self, type_class_or_function, is_return_type, storage_type):
 """:type: dict[str,TypeDefinition]"""
 
 
-def registerType(type_name, type_class_or_function, is_return_type=True, storage_type=None):
+def _registerType(type_name, type_class_or_function, is_return_type=True, storage_type=None):
```
jepebe: Good! :)

PR #1349, opened 2016-12-07
micro fix for debugging export problems

pgdr:


pgdr:
Self-merged, trivial and necessary fix in the debugging of export issues.

PR #1352, opened 2016-12-08
Python3 compatibility and types (Scheduled for 2.2)

pgdr:
We use `six.string_types` instead of `str` and `types.StringType`.  This is python2/3 compatible.

We also use `@six.metaclass` over `__metaclass__`, python2/3 compatible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ install:
     - conda update -q conda
     # Useful for debugging any issues with conda
     - conda info -a
-    - conda install pyqt=4 scipy=0.16.1 numpy pandas matplotlib
+    - conda install pyqt=4 scipy=0.16.1 numpy pandas matplotlib future
 
 env:
   global:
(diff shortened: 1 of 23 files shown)


pgdr:
The interesting thing to verify is that the following line does not throw `ImportError`:
```python
from builtins import ( bytes, int, list, range, str )
```

jepebe:
I think you should go for a `__repr__` in `BaseCClass` that contains something like:

```
def __repr__(self):
    return "{classname}(%s) {pointer}".format(classname=self.__class__.__name__, pointer=self._ad_str())
```

Then any child class can do something like this:

```
def __repr__(self): 
    return super(BaseCClass, self).__repr__() % "name='FOPT', value='9'"
```

pgdr:
I agree.  I will implement it before I add one more `__repr__`!

joakim-hove:
Will you rebase this to current master; and we will merge it.

joakim-hove:
Will you rebase this to current master; and we will merge it.

Issue #1235, opened 2016-09-22
Is `test_ecl_submit.py` used?


pgdr:
Is `make_run_path` ever called?

pgdr:
Closed by #1355.

PR #1355, opened 2016-12-08
removed unused test EclSubmitTest

pgdr:
the test contained no test methods

No test inherit from EclSubmitTest

Never is `make_run_path` run.


pgdr:
Clsoes #1235

PR #1358, opened 2016-12-12
Remove deprecated functionality (Scheduled for 2.2)

pgdr:
Do not merge!

This PR is _scheduled for merging_ after **branch 2.1** has been created.

The PR removes all functionality that has been deprecated since 1.9 and 2.0.  Updated many tests that used the deprecated API and removed tests that verify deprecation.

Implemented Ecl3DKW constructor (prior to this, we had to call a class method `.create` to obtain an instance.)  We should consider deprecating `Ecl3DKW.create` now to be consistent with its super class.


joakim-hove:
We will merge this quite soon after the 2.1 branch has been created.

pgdr:
I merged this as per Joakim's comment.

PR #1360, opened 2016-12-13
implemented missing function hook_manager_get_runpath_list

pgdr:
Implemented the missing prototype
```python
_get_runpath_list = EnkfPrototype("runpath_list_ref hook_manager_get_runpath_list(hook_manager)")
```

Added tests for HookManager and RunpathList from EnkfMain.


joakim-hove:
Yes - this makes sense.

pgdr:
I don't really know how to construct a sensible HookManager, so I only get an empty HookManager with an empty RunpathList.

Hence, the only thing I test is that I actually manage to get proper HookManager and RunpathList objects.

PR #1362, opened 2016-12-13
WIP/POC: implement stream in cwrap

pgdr:


python/python/cwrap/stream.py
```diff
@@ -0,0 +1,123 @@
...
+        super(LibcPrototype, self).__init__(LibcPrototype.lib, prototype, bind=bind)
+
+class Stream(BaseCClass):
+    """
+    Utility class to map a Python file handle <-> FILE* in C
+    """
+    TYPE_NAME = "stream"
+
+    _fopen  = LibcPrototype("void* fopen(char*, char*)")
+    _fread  = LibcPrototype("int   fread(void*, int, int, stream)")
```
joakim-hove: These `int` values should be  `size_t`.

python/python/cwrap/stream.py
```diff
@@ -0,0 +1,123 @@
...
+            self._fread(wrt, 1, size, self)
+            return wrt
+
+    def readline(self):
+        self._ass_open()
+        out = ''
+        while True:
+            ch = self.read(1)
+            out += ch
+            if ch == '\n':
```
joakim-hove: I guess Python has some platform dependent newline encoding somewhere.

joakim-hove:
Now - this is ambitious; but would of course be extremely cool if it worked out!

python/python/cwrap/stream.py
```diff
@@ -0,0 +1,123 @@
...
+            super(Stream, self).__init__(c_ptr)
+        except ValueError as e:
+            raise IOError('Could not load file "%s" in mode %s.' % (fname, mode))
+
+    def _ass_open(self):
+        if self.closed:
+            raise IOError('File "%s" is closed.' % self._fname)
+
+    def read(self, size=-1):
+        self._ass_open()
```
jepebe: Hahahaha!

python/python/cwrap/stream.py
```diff
@@ -0,0 +1,177 @@
...
+
+    _fclose = LibcPrototype("int   fclose (stream)", bind = True);
+    _fflush = LibcPrototype("int   fflush (stream)", bind = True);
+
+    def __init__(self, fname, mode = 'rb'):
+        c_ptr = self._fopen(fname, mode)
+        self._mode = mode
+        self._fname = fname
+
+        self.closed = False # A closed file cannot be used for further I/O
```
jepebe: Shouldn't this be prefixed with `_`and then have a read only property for closed?

python/python/cwrap/stream.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def seek(self, offset, whence):
+        """whence must be one of SEEK_SET, SEEK_CUR, or SEEK_END."""
+        self._assert_open()
+        if self._fseek(offset, whence):
+            raise IOError("Unable to seek.")
+
+    def _bytemode(self):
+        return 'b' in self._mode
+    def _textmode(self):
```
jepebe: Some space please.

python/python/cwrap/stream.py
```diff
@@ -0,0 +1,177 @@
...
+
+        if isinstance(np_arr, string_types):
+            arr = numpy.zeros(len(np_arr), dtype=numpy.character)
+            for i in range(len(np_arr)):
+                arr[i] = np_arr[i]
+            np_arr = arr
+        return self._fwrite(np_arr.ctypes.data, 1, len(np_arr), self)
+
+    def close(self):
+        print('%s.close()' % self)
```
jepebe: Debug?

python/python/cwrap/stream.py
```diff
@@ -0,0 +1,177 @@
...
+        if self._fseek(offset, whence):
+            raise IOError("Unable to seek.")
+
+    def _bytemode(self):
+        return 'b' in self._mode
+    def _textmode(self):
+        return 'b' not in self._mode
+
+    def _writable(self):
+        for m in 'wa+':
```
jepebe: `return any(m in self._mode for m in 'rb+')`

joakim-hove:
Possible alternative based on fd: http://stackoverflow.com/questions/23008895/is-it-possible-to-rescue-file-descriptor-from-file

pgdr:
Thanks, yes, this implementation needs to be thoroughly discussed.  Will the example you linked to have the seek synchronized?  One of the main issues is the problem of synchronizing the stream positions, and it doesn't seem like that example handles that.

python/python/cwrap/stream.py
pgdr:
```python
return any(m in self._mode for m in 'wa+')
```

PR #1364, opened 2016-12-14
Add seabed to geertsma kernel

kjellkongsvik:


pgdr:
No test with nonzero `seabed`?

kjellkongsvik:
@pgdr yes, should do that but hard to make that meaningful with synthetic data. But ok, can add one such

PR #1366, opened 2016-12-14
Potential bug?

arielalmendral:


libanalysis/src/std_enkf.c
```diff
@@ -325,7 +325,7 @@ bool std_enkf_get_bool( const void * arg, const char * var_name) {
       return module_data->use_EE;
     else if (strcmp(var_name , USE_GE_KEY_) == 0)
       return module_data->use_GE;
-    if (strcmp(var_name , ANALYSIS_SCALE_DATA_KEY_) == 0)
```
joakim-hove: Well; I would not say it is a bug - but the code would probably read better as:
```C
if (strcmp( var_name , USE_GE_KEY_) == 0)
   return module->use_GE;

if (strcmp( var_name , ANALYSIS_SCALE_DATA_KEY_) == 0)
   return module_date->analysis_scale_data;

if (strcmp( var_name , ...)

return false;
```
i.e. *only* using `if` and `return`.

arielalmendral: Agree, just got a bit confused. I close this one and make a new PR, keeping the old logic.

PR #1368, opened 2016-12-14
Fix bug in load case manually

pgdr:
`ValueModel` has no attribute `getActiveIteration`, must use `getValue()`.

`getValue()` is a string, so we try to cast to int.

_internal: ERT-1134_

### DIFF
--- a/python/python/ert_gui/tools/load_results/load_results_panel.py
+++ b/python/python/ert_gui/tools/load_results/load_results_panel.py
@@ -83,9 +83,11 @@ class LoadResultsPanel(QWidget):
         realizations = self._active_realizations_model.getActiveRealizationsMask()
         iteration = self._iterations_model.getValue()
         try:
+            if iteration is None:
+                iteration = ''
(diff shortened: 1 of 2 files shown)


python/python/ert_gui/tools/load_results/load_results_panel.py
```diff
@@ -81,9 +81,14 @@ def load(self):
         all_cases = self._case_model.getAllItems()
         selected_case  = all_cases[self._case_combo.currentIndex()]
         realizations = self._active_realizations_model.getActiveRealizationsMask()
-        iteration = self._iterations_model.getActiveIteration()
-
+        iteration = self._iterations_model.getValue()
+        try:
+            iteration = int(iteration)
+        except ValueError as e:
+            print(e)
```
jepebe: Debug? Either remove or make more informative.

pgdr: Fixed.

PR #1374, opened 2016-12-15
printing when loading forward models

pgdr:


libenkf/src/enkf_main.c
```diff
@@ -3109,11 +3109,13 @@ void enkf_main_load_from_forward_model_with_fs(enkf_main_type * enkf_main, int i
 
   int iens = 0;
   for (; iens < ens_size; ++iens) {
+      printf("\tloading %d (realization %d/%d) ", iens, (1+iens), ens_size);
     result[iens] = 0;
     arg_pack_type * arg_pack = arg_pack_alloc();
     arg_list[iens] = arg_pack;
 
     if (bool_vector_iget(iactive, iens)) {
+      printf("... ");
```
joakim-hove: This should call `flush(stdout)` to have any value.

PR #1375, opened 2016-12-15
repr for bool_vector and enkf_node

pgdr:
during debugging I needed to know some states of `BoolVector` and `EnkfNode`, so I implemented `__repr__`.

They might be good to have later as well.


python/python/ert/util/bool_vector.py
```diff
@@ -143,3 +143,23 @@ def createFromList(cls, size, source_list):
     def createActiveList(self):
         """ @rtype: ert.util.IntVector """
         return self._active_list(self)
+
+    def _tostr(self, arr = None):
```
joakim-hove: Do not need the optional `arr`argument?

PR #1378, opened 2016-12-16
Added __setitem__ and __getitem__ for neg, and repr

pgdr:
* vector_template now supports setting negative indices and slices.
* testing the repr of BoolVector

### DIFF
--- a/python/python/ert/util/vector_template.py
+++ b/python/python/ert/util/vector_template.py
@@ -205,13 +205,14 @@ class VectorTemplate(BaseCClass):
         """
         if isinstance(index, IntType):
             length = len(self)
-            if index < 0:
-                index += length
+            idx = index
+            if idx < 0:
+                idx += length
 
-            if index < 0 or index >= length:
-                raise IndexError("Index must be in range %d <= %d < %d" % (0, index, length))
+            if 0 <= idx < length:
+                return self._iget(idx)
             else:
-                return self._iget(index)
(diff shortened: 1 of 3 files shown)


python/python/ert/util/vector_template.py
```diff
@@ -205,26 +205,34 @@ def __getitem__(self, index):
...
         """
+        ls = len(self)
         if isinstance(index, IntType):
-            self._iset(index, value)
+            idx = index
+            if idx < 0:
+                idx += ls
+            self._iset(idx, value)
+        elif isinstance( index, slice ):
+            for i in range(*index.indices(ls)):
```
joakim-hove: Is this sliced assignment 100% according to standard Python?

python/tests/core/util/test_vectors.py
```diff
@@ -104,6 +104,34 @@ def test_vector_operations_with_exceptions(self):
         with self.assertRaises(TypeError):
             iv1 *= dv1
 
+    def test_setitem_getitem(self):
+        primes = [2,3,5,7,11,13,17,19]
+        primep = [i in primes for i in range(20)]
+        b = BoolVector(initial_size = 20)
```
joakim-hove: This `BoolVector` should be initialized with `initial_value = False`. Now I fear you are not really testing the `__setitem__()` functionality.

joakim-hove: Ok - I was wrong again, the default is fslse.

PR #1384, opened 2016-12-21
Sum vector issues

pgdr:
Added `__iter__` for `EclSum` (iterate over keys so we can `get_values(keys)`).  Implement `__repr__` for `EclSumVector` and `VectorTemplate`.

But _most importantly_ this fixes two bugs:

* The first commit fixes getting `unit` from `EclSumVector`.
* The second commit fixes `TimeVector` slicing.

The third commit adds testing for `Double`, `Int`, `Bool`, and `Time`-`Vector`.

### DIFF
--- a/python/python/ert/ecl/ecl_sum.py
+++ b/python/python/ert/ecl/ecl_sum.py
@@ -430,6 +430,9 @@ class EclSum(BaseCClass):
         if not key in self:
             raise KeyError("The summary key:%s was not recognized" % key)
 
+    def __iter__(self):
+        return iter(self.keys())
+
     def __getitem__(self , key):
         """
         Implements [] operator - @key should be a summary key.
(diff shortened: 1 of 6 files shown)


pgdr:
I think I've replicated the `dict` behavior.  For a mapping `m: X → Y`, then `iter(m)` has type `X` and `m[·]` has type `Y`.

```python
>>> d = {'a':1, 'b':2}
>>> list(iter(d))
['a', 'b']
>>> for k in d:
...:    print(d[k])
...:
1
2
```

pgdr:
Should I remove `__iter__`?  That's no problem, I just wanted to use it in Håvard's exportCSV script.

joakim-hove:
The argument about `__dict__` was good.

pgdr:
I added some tests, I'll merge this when it greens.

PR #1385, opened 2016-12-21
Setting correct working directory, and build target dependency for sphinx-apidoc / html generation

thorvaldj:


joakim-hove:
Great!!

PR #1387, opened 2016-12-23
Remove `cstring_obj`

pgdr:
Replaced `cstring_obj` with `char*`

1.  Added `__repr__` to EnKFMain.
2.  moved get_init_file from enkf_main to config
3.  Removed the type `cstring_obj` and use `char*` instead.


jepebe:
If I remember correctly this was made for a reason.

jepebe:
The string that is returned needs to be freed. The default implementation for `char*` assumes `const` or handled somewhere else.

pgdr:
Thanks, I see that that is a valid objection.

Then I wonder if I should keep a reference in the C struct and free it when `EnKFMain` is freed.

It's annoying to have a type (quite non-standard compared with the rest of ert python bindings) for just one function.  The reason I made this PR was because the `cstring`  test is flaky as well, meaning, maybe we're touching _undefined behavior area._

What do you think?

joakim-hove:
> Then I wonder if I should keep a reference in the C struct and free it when EnKFMain is freed.

I think that is a reasonable approach; it should be a container of some sort - e.g. `stringlist`.

>It's annoying to have a type (quite non-standard compared with the rest of ert python bindings) for just one function. 

Absolutely agreed.

jepebe:
I agree that an alternate implementation is preferred. On the Python side `char*` will be copied into a Python string and the pointer is dropped, so you are free to do whatever you want on the `C` side.

libenkf/src/enkf_config_node.c
```diff
@@ -565,6 +566,9 @@ void enkf_config_node_free(enkf_config_node_type * node) {
   free(node->key);
   stringlist_free(node->obs_keys);
 
+  if (node->init_file_abs_path != NULL)
```
joakim-hove: The code already has many of these checks - but they are actually not necessary. Calling `free( NULL )` is a well defined no-op: http://stackoverflow.com/questions/1938735/does-freeptr-where-ptr-is-null-corrupt-memory

libenkf/src/enkf_config_node.c
```diff
@@ -603,6 +607,43 @@ const char * enkf_config_node_get_enkf_infile( const enkf_config_node_type * con
...
+  bool forward_init                = enkf_config_node_use_forward_init(config_node);
+
+  if (forward_init) {
+    path_fmt_type * runpath_fmt = model_config_get_runpath_fmt(model_config);
+    runpath = path_fmt_alloc_path(runpath_fmt , false , 0, 0);  /* Replace first %d with iens, if a second %d replace with iter */
+    path_to_init_file = enkf_config_node_alloc_initfile(config_node, runpath, 0);
+  } else
+    path_to_init_file = enkf_config_node_alloc_initfile(config_node, NULL, 0);
+
+  if (path_to_init_file)
```
joakim-hove: The existence check can be directly on the `path_to_init_file` variable, not necessary to first create the absolute variety.

libenkf/src/enkf_config_node.c
```diff
@@ -603,6 +607,43 @@ const char * enkf_config_node_get_enkf_infile( const enkf_config_node_type * con
...
+    abs_path_to_init_file = util_alloc_abs_path(path_to_init_file);
+
+  if (abs_path_to_init_file && !util_file_exists(abs_path_to_init_file)) {
+    free(abs_path_to_init_file);
+    abs_path_to_init_file = NULL;
+  }
+
+  if (runpath)
```
joakim-hove: These can just be: 
```C
free( runpath );
free( path_to_init_file );
```

libenkf/src/enkf_config_node.c
```diff
@@ -603,6 +607,43 @@ const char * enkf_config_node_get_enkf_infile( const enkf_config_node_type * con
   return path_fmt_get_fmt( config_node->enkf_infile_fmt );
 }
 
+
+const char * enkf_config_node_get_init_file(enkf_config_node_type * config_node, const enkf_main_type * enkf_main ) {
```
joakim-hove: This function should take a `model_config_type` as input - and not a full oversized `enkf_main` instance - I am honestly surprised it even compiles like this!?

pgdr:
@joakim-hove thanks for the comments!  This is still WIP, my apologies for not adding WIP to the title (it's done now).

joakim-hove:
> This is still WIP, my apologies for not adding WIP to the title (it's done now).

No worries - this will be a good contribution when it is in.

pgdr:
ready for review

libenkf/src/enkf_config_node.c
```diff
@@ -603,6 +607,35 @@ const char * enkf_config_node_get_enkf_infile( const enkf_config_node_type * con
   return path_fmt_get_fmt( config_node->enkf_infile_fmt );
 }
 
+
+const char * enkf_config_node_get_init_file(enkf_config_node_type * config_node, const path_fmt_type * runpath_fmt) {
```
joakim-hove: Disclaimer: I used quite long time to disentange / recap this myself. 

The name `init_file` - which was also used in the previous code - is unfortunate, because it hints at initialization, which is an important task which applies to all the ert node types. However - this is a very specialized use case; only applicable when exporting a field - e.g. `PORO` or `PERMX`. Ert ony internalizes the active cells of a field, by using the filename given here the export routine gets access to a "templatee/fill in" field which can be used for the inactive cells, the same field is used for all realizations. 

Suggestions:

1. This function is renamed to something like `enkf_config_node_get_FIELD_fill_file()`
2. The `getFieldInitFile()` is completely removed from the `EnkfMain` class - instead the two call sites: [here](https://github.com/Ensembles/ert/blob/master/python/python/ert_gui/shell/export.py#L59) and [here](https://github.com/Ensembles/ert/blob/master/python/python/ert_gui/tools/export/export_model.py#L47) use the config node implementation directly.

pgdr:
@joakim-hove all comments addressed.  ready for re-review.

PR #1389, opened 2016-12-29
Minor field config refactor, added type FieldExportEnum

pgdr:
The `FieldConfig.get_type()`, with return type `int`, was prone to return `-1`

The function now returns a new type, `FieldTypeEnum`, there is a corresponding `c` enum.

Added a new type `UNKNOWN=4` and initializes the `field_config` objects with `type=UNKNOWN`.

### DIFF
--- a/libecl/src/ecl_kw.c
+++ b/libecl/src/ecl_kw.c
@@ -2518,41 +2518,6 @@ bool ecl_kw_is_kw_file(fortio_type * fortio) {
 }
 
 
-
-
-
-bool ecl_kw_is_grdecl_file(FILE * stream) {
-  const long int init_pos = util_ftell(stream);
-  bool grdecl_file;
-  bool at_eof = false;
-  util_fskip_chars(stream ,  " \r\n\t"  , &at_eof);  /* Skipping intial space */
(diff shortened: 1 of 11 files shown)


libecl/src/ecl_kw.c
```diff
@@ -2518,36 +2518,39 @@ bool ecl_kw_is_kw_file(fortio_type * fortio) {
 }
 
 
-
-
+static bool ecl_kw_is_grdecl_file_rewind(FILE * stream, const long int pos, bool ret_val) {
```
joakim-hove: Use `offset_type` for file offsets instead of naked `long int`. See discussion [here](https://github.com/Ensembles/ert/blob/master/libert_util/src/util_lfs.c)

joakim-hove: Since this function is so *very* internal, I think it should not even get the `ecl_kw_`prefix - I would prefer something like:
`rewind_and_return()`?

pgdr:
@joakim-hove I have addressed your comments by deleting the function I wrote.

I also deleted the existing function `ecl_kw_is_grdecl_file` and instead use the function `ecl_kw_grdecl_fseek_next_kw`.  I don't know if this is good enough, but no tests fail.

I added a test that tests `guessFiletype` by first exposing that function to python.

PR #1390, opened 2017-01-02
Removing outdated keywords from user documentation

thorvaldj:


docs/user/keywords/index.rst
```diff
@@ -106,9 +106,7 @@ Keyword name                                                        	Required by
 :ref:`MAX_SUBMIT <max_submit>` 						NO 					2 				How many times should the queue system retry a simulation. 
 :ref:`MIN_REALIZATIONS <min_realizations>` 				NO 					0 				Set the number of minimum reservoir realizations to run before long running realizations are stopped. Keyword STOP_LONG_RUNNING must be set to TRUE when MIN_REALIZATIONS are set. 
 :ref:`NUM_REALIZATIONS <num_realizations>` 				YES 									Set the number of reservoir realizations to use. 
-:ref:`OBS_CONFIG <obs_config>` 						NO 									File specifying observations with uncertainties. 
-:ref:`PLOT_SETTINGS <plot_driver>` 					NO 					  				Possibility to configure some aspects of plotting.
```
joakim-hove: `PLOT_SETTINGS` is the new to configure plot settings, and that should definitely be retained. This should actually be a "super setting" which should apply to all plot related settings.

thorvaldj: True.. I went for full removal of PLOT_*. Should be back again now.

PR #1391, opened 2017-01-02
Added settings object for update settings.

joakim-hove:


pgdr:
seems a `plot_settings` test fails:

`263 - python.tests.ert.enkf.plot.plot_settings (Failed)`

joakim-hove:
> seems a plot_settings test fails:

Yes - and that was actually more difficult than I thought to fix; working on it.

docs/user/keywords/index.rst
```diff
@@ -1783,6 +1773,47 @@ The two keywords SETENV and UPDATE_PATH can be used to manipulate the Unix envir
...
+
+	Including outliers in the Smoother algorithm can dramatically
+	increase the coupling between the ensemble members. It is
+	therefore important to filter out these outlier data prior to
+	data assimilation. An observation, \textstyle d^o_i, will be
+	classified as an outlier if
+
+	::
+
+		|d^o_i - \bar{d}_i| > \mathrm{ENKF\_ALPHA} \left(s_{d_i} + \sigma_{d^o_i}\right),
```
pgdr: no need to change, but `\mathrm` can (and should) in most cases be simply `\text` or `\textnormal`.

`\mathrm` means that you want to go to roman font, which you should rarely (if ever) do.  `\text` simply switches to text mode.

libenkf/src/analysis_config.c
```diff
@@ -38,6 +39,10 @@
 #include <ert/enkf/analysis_iter_config.h>
 
 
+#define UPDATE_OVERLAP_KEY      "OVERLAP_LIMIT"
```
pgdr: Why is alpha removed as part of the name here but not in the API?

joakim-hove:
> The only comment I have is that the config_settings should have been called a multityped_map or whatever and been placed in util.

Yes - that might be; it is quite related to config input though?

> The next time I need a map in ert, I'm not going to use config_settings_type.

Why? Because it is located in a unlogical place, or because you don't like this particular implementation?

pgdr:
> Why? Because it is located in a unlogical place, or because you don't like this particular implementation?

Nono, I like the implementation!  It's just that I will look in `util` for something with `map` in its name.

joakim-hove:
> Nono, I like the implementation! It's just that I will look in util for something with map in its name.

OK good - I understand; then I suggest we keep it here for now (it is reasonably tighly bound the config system afterall), and then we might extract parts of it to util later.

pgdr:
Agreed.

PR #1395, opened 2017-01-05
Get (x, y) pos from surface

kjellkongsvik:


libgeometry/src/geo_surface.c
```diff
@@ -375,6 +375,12 @@ geo_pointset_type * geo_surface_get_pointset( const geo_surface_type * surface )
   return surface->pointset;
 }
 
+void geo_surface_get_xy( const geo_surface_type* surface, int index, double* x, double* y) {
```
joakim-hove: Use the name `geo_surface_iget_xy()` - there is a quite strong convention that index based lookup gets the `iget(..)` name.

joakim-hove:
Apart from the minor name comment this is ready to be merged.

joakim-hove:
:+1:

PR #1397, opened 2017-01-10
Apply plot style to other plots

thorvaldj:


python/python/ert_gui/ertwidgets/checklist.py
```diff
@@ -111,9 +129,20 @@ def filterList(self, filter):
...
+        """
+        for index in range(0, self._list.count()):
+            item = self._list.item(index)
+            if not item.isHidden():
+                self._model.selectValue(str(item.text()))
 
     def uncheckAll(self):
+        """
+        Unchecks all items in the list, visible or not
+        :return:
```
jepebe: Should not be here

python/python/ert_gui/tools/plot/filterable_kw_list_model.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class FilterableKwListModel(SelectableListModel):
+    """
+    Adds ERT - plotting keyword specific filtering functionality to the general SelectableListModel
+    """
+    def __init__(self, ert, selectable_keys):
+        SelectableListModel.__init__(self, selectable_keys)
+        self.__ert = ert
+        self.__show_summary_keys = True
```
jepebe: We generally do not use double underscore prefix

python/python/ert_gui/tools/plot/widgets/copy_style_to_dialog.py
```diff
@@ -0,0 +1,62 @@
...
+        filter_popup_button.clicked.connect(self.__filter_popup.show)
+
+        self._list_model = FilterableKwListModel(self._ert, selectable_keys)
+        self._list_model.unselectAll()
+
+        self._cl = CheckList(self._list_model, custom_filter_button=filter_popup_button)
+
+        layout.addWidget(self._cl)
+
+        _apply_button = QPushButton("Apply")
```
jepebe: Single underscore without `self.` is unnecessary.

PR #1401, opened 2017-01-18
Make ies advanced option only

thorvaldj:


python/python/ert_gui/simulation/simulation_panel.py
```diff
@@ -69,18 +69,34 @@ def addSimulationConfigPanel(self, panel):
...
         return []
 
+    def toggleAdvancedOptions(self, show_advanced):
+        current_model = self.getCurrentSimulationModel()
+
+        self._simulation_mode_combo.clear()
+        for model, panel in self._simulation_widgets.iteritems():
+            if show_advanced:
+                self._simulation_mode_combo.addItem(str(model), model)
+            elif not panel.isAdvancedOption:
```
jepebe: `if show_advanced or not panel.isAdvancedOption:` ?

python/python/ert_gui/simulation/simulation_config_panel.py
```diff
@@ -6,10 +6,15 @@ class SimulationConfigPanel(QWidget):
...
 
-    def __init__(self, simulation_model):
+    def __init__(self, simulation_model, advanced_option=False):
         QWidget.__init__(self)
         self.setContentsMargins(10, 10, 10, 10)
         self.__simulation_model = simulation_model
+        self._advanced_option = advanced_option
+
+    @property
+    def isAdvancedOption(self):
```
jepebe: A `property` should not have camelCase since it is not a function.

PR #1406, opened 2017-01-27
Ability to skip keywords in binary file / support for type 'C010'

joakim-hove:
Keywords with the data type 'C010' have been encountered - a guess-based implementation has been added.


libecl/src/ecl_kw.c
```diff
@@ -1240,9 +1259,11 @@ bool ecl_kw_fskip_data__( ecl_type_enum ecl_type , const int element_count , for
       const int block_count = element_count / blocksize + (element_count % blocksize == 0 ? 0 : 1);
 
       int element_size = ecl_util_get_sizeof_ctype(ecl_type );
-      if(ecl_type == ECL_CHAR_TYPE || ecl_type == ECL_MESS_TYPE) {
```
pgdr: what happened with type `ECL_MESS`?

pgdr:
A pleasure to review!

If the tests pass the StatoilData testsuite, go ahead and merge.  One question raised, but no need to answer it if you understand the situation and guarantee its correctness. (the removal of `ECL_MESS_TYPE` from an `if`).

libecl/src/ecl_kw.c
joakim-hove: Well - the element count of keywords of type `ECL_MESS`is *always* zero - i.e. there is no attached data; so I skipped it.

PR #1407, opened 2017-01-30
Output correct unit string when writing GRID/EGRID

joakim-hove:
Previously the GRID/EGRID files were written with the keyword:
```txt
GRIDUNITS   2  'CHAR' 
'METRES  '  '        ' 
```
irrespective of the units used, this will now correctly output `FEET` in field units.


joakim-hove:
Test added :-)

PR #1418, opened 2017-02-17
JSON output of forward models

markusdregi:
We are transitioning to Json for the configuration file for JobMananger. We have added functionality for specifying how Null is to be presented when written, a first step towards Json compatibility.

### DIFF
--- a/python/tests/ert/job_queue/test_forward_model_formatted_print.py
+++ b/python/tests/ert/job_queue/test_forward_model_formatted_print.py
@@ -1,0 +1,402 @@
+import os.path
+import json
+
+from ert.test import TestAreaContext, ExtendedTestCase
+from ert.util import SubstitutionList
+from ert.job_queue.forward_model import ForwardModel
+from ert.job_queue.ext_job import ExtJob
+from ert.job_queue.ext_joblist import ExtJoblist
+
+#
+# Data for testing
+#
+joblist = [
+            {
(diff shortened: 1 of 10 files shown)


markusdregi:
The forward model will now write itself in JSON to file in addition to the current python format. This is done to avoid the somewhat unexpected behaviour loading the data as a python module can result in.

The next step is to entirely abandon the python format. The code has been written such that programatically this is a very easy job. The code that is to be removed is listed in ERT-1173.

python/tests/ert/job_queue/test_forward_model_formatted_print.py
```diff
@@ -0,0 +1,402 @@
...
+            if not os.path.isabs(job["executable"]):
+                job["executable"] = os.path.join(
+                        os.getcwd(),
+                        job["executable"]
+                        )
+
+    def validate_ext_job(self, ext_job, ext_job_config):
+        zero_if_none = lambda x: 0 if x is None else x
+
+        self.assertEqual(
```
joakim-hove: Is this a new style syntax? Personally I would prefer the old school:

```python
self.assertEqual( ext_job.name() , default_name_if_none( ext_job_config["name"])
```

markusdregi: If put on one line it would definitively be more than 80 characters long. And that is not old school..

PR #1423, opened 2017-03-01
Test Area permissions

markusdregi:


libert_util/src/test_work_area.c
```diff
@@ -137,8 +137,9 @@ static test_work_area_type * test_work_area_alloc__(const char * prefix , const
       work_area->original_cwd = util_alloc_cwd();
       work_area->cwd = test_cwd;
       work_area->change_dir = change_dir;
-      if (change_dir)
-        util_chdir( work_area->cwd );
+      if (change_dir )
+        if(util_chdir( work_area->cwd ) == -1)
```
joakim-hove: I guess this should be: `if (util_chdir( work_area->cwd ) != 0)`?

libert_util/src/test_work_area.c
```diff
@@ -137,8 +137,9 @@ static test_work_area_type * test_work_area_alloc__(const char * prefix , const
...
+            util_abort("Failed to move into temporary directory: %s", test_cwd);
```
joakim-hove: The convention for these crashes is to pass the current function name as first argument:
```C
util_abort("%s: Failed to change to temporary directory: %s", __func__ , test_cwd);
```

libert_util/src/test_work_area.c
markusdregi: The return value is guaranteed to be either 0 or -1, but I'll happily change it to your suggestion..

PR #1424, opened 2017-03-01
Issue with `char* ecl_grid_get_name( ecl_grid )` returning nullptr

pgdr:
Improved `EclGrid.getName` function, which made `EclGrid.__repr__(·)` "hickup".

In the case that a grid is synthetically constructed, `EclGrid.getName()` returned `None`.  Now we guarantee that `EclGrid.getName()` returns a string.

### DIFF
--- a/python/tests/core/ecl/test_grid.py
+++ b/python/tests/core/ecl/test_grid.py
@@ -85,10 +85,22 @@ class GridTest(ExtendedTestCase):
         actnum[1] = 0
         grid = EclGrid.createRectangular( (10,20,30) , (1,1,1) , actnum = actnum)
         self.assertEqual( grid.getNumActive( ) , 30*20*10 - 2)
-    
-    
-    
-    
+
+    def test_repr_and_name(self):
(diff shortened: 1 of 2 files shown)


python/tests/core/ecl/test_grid.py
```diff
@@ -85,10 +85,23 @@ def test_create(self):
...
-    
+
+    def test_repr_and_name(self):
+        grid = EclGrid.createRectangular((2,2,2), (10,10,10), actnum=[0,0,0,0,1,1,1,1])
+        pfx = 'EclGrid('
+        rep = repr(grid)
+        self.assertEqual(pfx, rep[:len(pfx)])
+        self.assertEqual(type(rep), type(''))
+        self.assertEqual(type(grid.getName()), type(''))
+        print(rep)
```
joakim-hove: I guess the `print(rep)` should go?

pgdr: yes, thanks

pgdr: done, force-pushed amended commit

PR #1425, opened 2017-03-01
WIP removed some uses of CFILE

pgdr:
The goal with this WIP PR is to have a conversation regarding the future of CFILE, streams, file handles and descriptors, and python 3 compatibility.

Today, the CFILE implementation and strategy is not Python 3 compatible, and not possible to make python 3 compatible (due to the removal of `FILE *fp = PyFile_AsFile(p);` from Python 3).

We should agree on a strategy to get rid of the use of `PyFile`.

Kjør debatt.


joakim-hove:
Good: I want proof-of-concept of this approach:https://github.com/Ensembles/ert/pull/1362#issuecomment-275952617

Issue #1431, opened 2017-03-07
Small issue with python 3.4


dr-robertk:
The file python/python/ert/ecl/ecl_rft.py contains a tab in line 91 which prevents from the package from building with python 3.4. Replacing the tab with white spaces fixes the issue.

joakim-hove:
OK - cool; we will fix that. Be aware that there are more fundamental issues than space/tabs which prevents fully functional Python3 support - but we are on it.

pgdr:
That's freakin' embarrassing!

#1432 fixes this.

PR #1432, opened 2017-03-07
replaced tab with `    `

pgdr:


pgdr:
Fixes #1431.

joakim-hove:
Wow - you are fast.

PR #1439, opened 2017-03-10
New eclipse data type 

markusdregi:
The first step towards supporting variable string length eclipse keywords (aka. Cxxx). The new data type ecl_data_type is a struct encapsulating the old ecl_type_enum, describing the type of the data, together with the length (that now can vary).

While propagating the new data type, the functionality of the type, both new and old, is moved to ecl_type.c.


libecl/include/ert/ecl/ecl_type.h
```diff
@@ -0,0 +1,123 @@
...
+
+#define ECL_TYPE_ENUM_SIZE 6
+
+
+struct ecl_type_struct {
+    const ecl_type_enum type;
+
+    // In case of ECL_STRING_TYPE:
+    // length denotes the number of characters in the string
+    const size_t length;
```
joakim-hove: If we introudcue this structure I think we should use it consistently - i.e. the `sizeof_ctype` field should be removed from the `struct ecl_kw_struct` - and:

1. `length` should be renamed `element_size` or something like that.
2. `element_size` should be correctly initialized for all the types, i.e.

```
#define ECL_INT (ecl_data_type) {.type = ECL_INT_TYPE, .element_size = 4}
```

joakim-hove:
Hmmm - where is the `tests/ecl_type_test.c` file?:smile:

libecl/include/ert/ecl/ecl_type.h
markusdregi: I agree!

libecl/src/ecl_type.c
```diff
@@ -0,0 +1,204 @@
...
+}
+
+void ecl_type_free(ecl_data_type * ecl_type) {
+    if(ecl_type == NULL)
+        return;
+
+    free(ecl_type);
+}
+
+ecl_data_type ecl_type_get_data_type(const ecl_type_enum type, const size_t element_size) {
```
joakim-hove: This is function creates a *new* object - i.e. it is *not* a *get* function. I would suggest `ecl_type / ecl_type_make / ecl_type_create / .... `

libecl/src/ecl_type.c
```diff
@@ -0,0 +1,204 @@
...
+
+ecl_data_type * ecl_type_alloc_copy(const ecl_data_type * src_type) {
+    ecl_data_type * ecl_type;
+    ecl_type = util_malloc(ecl_type_get_sizeof_ctype(src_type));
+    memcpy(ecl_type, src_type, ecl_type_get_sizeof_ctype(src_type));
+    return ecl_type;
+}
+
+void ecl_type_free(ecl_data_type * ecl_type) {
+    if(ecl_type == NULL)
```
joakim-hove: Calling `free( NULL `)` is a safe no-op.

libecl/src/ecl_type.c
```diff
@@ -0,0 +1,204 @@
...
+*/
+#define ECL_TYPE_NAME_CHAR     "CHAR"
+#define ECL_TYPE_NAME_C010     "C010"
+#define ECL_TYPE_NAME_FLOAT    "REAL"
+#define ECL_TYPE_NAME_INT      "INTE"
+#define ECL_TYPE_NAME_DOUBLE   "DOUB"
+#define ECL_TYPE_NAME_BOOL     "LOGI"
+#define ECL_TYPE_NAME_MESSAGE  "MESS"
+
+ecl_data_type * ecl_type_alloc_copy(const ecl_data_type * src_type) {
```
joakim-hove: As I see it the plan here is to use value semantics - with a publicly available `struct` definition. I would therefor suggest we just avoid implementing pointer specific functions like `ecl_type_alloc_copy( )` and `ecl_type_free( )`.

libecl/src/ecl_type.c
```diff
@@ -0,0 +1,204 @@
...
+
+    if(ecl_type->element_size != element_size)
+        util_abort(
+                "%s: element_size mismatch for type %d, was: %d, expected: %d\n",
+                __func__, type, element_size, ecl_type->element_size);
+
+    return *ecl_type;
+}
+
+ecl_data_type ecl_type_get_data_type_from_type(const ecl_type_enum type) {
```
joakim-hove: It is not *get* but rather *make*.

libecl/src/ecl_type.c
```diff
@@ -0,0 +1,204 @@
...
+    ecl_data_type * ecl_type = NULL;
```
joakim-hove: No reason to go through the pointer, addressof and dereferencing - this should cut it:

switch(type) {
   case(ECL_CHAR_TYPE) : return ECL_CHAR;
   case(ECL_INT_TYPE) :  return ECL_INT;
   ..
```.

File: libecl/include/ert/ecl/ecl_kw.h [hunk 7]
```diff
@@ -61,12 +62,14 @@ extern "C" {
...
   bool           ecl_kw_fread_realloc_data(ecl_kw_type *ecl_kw, fortio_type *fortio);
-  ecl_type_enum  ecl_kw_get_type(const ecl_kw_type *);
+  ecl_data_type  ecl_kw_get_data_type(const ecl_kw_type *);
+  size_t         ecl_kw_get_sizeof_ctype(const ecl_kw_type *);
   const char   * ecl_kw_get_header8(const ecl_kw_type *);
   const char   * ecl_kw_get_header(const ecl_kw_type * ecl_kw );
   ecl_kw_type  * ecl_kw_alloc_empty(void);
   ecl_read_status_enum ecl_kw_fread_header(ecl_kw_type *, fortio_type *);
   void           ecl_kw_set_header_name(ecl_kw_type * , const char * );
+  void           ecl_kw_set_data_type(ecl_kw_type * , ecl_data_type);
```
joakim-hove: Hmmm - should not be necessary with such a `set` function?

markusdregi: I agree

libecl/include/ert/ecl/ecl_type.h
```diff
@@ -0,0 +1,79 @@
...
+
+typedef struct ecl_type_struct ecl_data_type;
+
+
+#define ECL_CHAR (ecl_data_type) {.type = ECL_CHAR_TYPE, .element_size = 8+1}
+#define ECL_INT (ecl_data_type) {.type = ECL_INT_TYPE, .element_size = sizeof(int)/sizeof(char)}
+#define ECL_FLOAT (ecl_data_type) {.type = ECL_FLOAT_TYPE, .element_size = sizeof(float)/sizeof(char)}
+#define ECL_DOUBLE (ecl_data_type) {.type = ECL_DOUBLE_TYPE, .element_size = sizeof(double)/sizeof(char)}
+#define ECL_BOOL (ecl_data_type) {.type = ECL_BOOL_TYPE, .element_size = sizeof(int)/sizeof(char)}
+#define ECL_MESS (ecl_data_type) {.type = ECL_MESS_TYPE, .element_size = 1}
```
joakim-hove: `element_size = 0`for `ECL_MESS`

libecl/include/ert/ecl/ecl_type.h
```diff
@@ -0,0 +1,79 @@
...
+
+    // Denotes the number of charachters used to represent
+    // the data type in *MEMORY*
+    const size_t element_size;
+};
+
+typedef struct ecl_type_struct ecl_data_type;
+
+
+#define ECL_CHAR (ecl_data_type) {.type = ECL_CHAR_TYPE, .element_size = 8+1}
```
joakim-hove: `sizeof(char)` is equal to 1 - by definition.

libecl/src/ecl_kw_grdecl.c
```diff
@@ -677,3 +677,11 @@ void ecl_kw_fprintf_grdecl(const ecl_kw_type * ecl_kw , FILE * stream) {
   ecl_kw_fprintf_grdecl__(ecl_kw , NULL , stream );
 }
 
+/**
+ *
+ * Functions only to be used by the *PYTHON* prototype!
+ *
+ */
+ecl_kw_type * python_ecl_kw_fscanf_alloc_grdecl_dynamic__( FILE * stream , const char * kw , bool strict , const ecl_data_type * data_type) {
```
joakim-hove: Make the `pthon` a suffix - i.e. `ecl_kw_fscanf_alloc_grdecl_dynamic_python__( )` - what a name!

libecl/src/ecl_type.c
```diff
@@ -0,0 +1,261 @@
...
+  else {
+    util_abort("%s: unrecognized type name:%s \n",__func__ , type_name);
+    return ECL_INT; /* Dummy */
+  }
+}
+
+
+int ecl_type_get_sizeof_ctype_fortio(const ecl_data_type ecl_type) {
+  if(ecl_type_is_char(ecl_type) || ecl_type_is_C010(ecl_type))
+      return (ecl_type.element_size - 1) * sizeof(char);
```
joakim-hove: `sizeof(char)` *is* 1.

libecl/src/ecl_type.c
```diff
@@ -0,0 +1,261 @@
...
+    return (ecl_type.type == ECL_C010_TYPE);
+}
+
+
+/**
+ *
+ * Functions only to be used by the *PYTHON* prototype!
+ *
+ */
+
```
joakim-hove: So many python functions they could get a file of their own.

joakim-hove:
Can we create some wrapping magic in Python so the old enum based way will continue to work, with a big fat warning?

markusdregi:
@joakim-hove: All of your suggestions are implemented

libecl/src/ecl_type.c
```diff
@@ -0,0 +1,172 @@
...
+ecl_type_enum ecl_type_get_type(const ecl_data_type ecl_type) {
+    return ecl_type.type;
+}
+
+size_t ecl_type_get_element_size(const ecl_data_type ecl_type) {
+    return ecl_type.element_size;
+}
+
+const char * ecl_type_get_type_name(const ecl_data_type ecl_type) {
```
joakim-hove: Can I have the slightly shorter: `ecl_type_get_name( )`?

joakim-hove:
OK - this is ready to be merged, but it will also affect opm (more than I thought ...) - so the merging must be coordinated.

See: https://github.com/OPM/opm-output/pull/176

PR #1441, opened 2017-03-10
Minor fortio fixes

pgdr:
1st commit: Context manager `__exit__` should return True if no exception occurred.

2nd commit: linting

### DIFF
--- a/python/python/ert/ecl/fortio.py
+++ b/python/python/ert/ecl/fortio.py
@@ -184,7 +184,7 @@ class FortIOContextManager(object):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.__fortio.close()
-        return False
+        return exc_type is not None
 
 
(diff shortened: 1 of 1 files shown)


python/python/ert/ecl/fortio.py
```diff
@@ -184,7 +184,7 @@ def __enter__(self):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.__fortio.close()
-        return False
+        return exc_type is not None
```
joakim-hove: I think we have that one wrong in many places ....

pgdr: That could be, do we have that many context managers?

The most important part is to _not_ return `True` always, I don't know exactly what the bad consequences are of always returning `False`, but I think it means that _"something bad happened and we didn't deal with it"_.

PR #1446, opened 2017-03-21
Prints message regarding MAX_SUBMIT

markusdregi:
When a node fails we should inform the user that this failure counts towards the specified MAX_SUBMIT.

### DIFF
--- a/libjob_queue/src/lsf_driver.c
+++ b/libjob_queue/src/lsf_driver.c
@@ -860,8 +860,8 @@ static void lsf_driver_node_failure(lsf_driver_type * driver, long lsf_job_id) {
     fprintf(stderr, "%s blacklisting nodes %s.\n", __func__, hostnames);
 
     fprintf(stderr,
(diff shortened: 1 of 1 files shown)


libjob_queue/src/lsf_driver.c
```diff
@@ -859,6 +859,15 @@ static void lsf_driver_node_failure(lsf_driver_type * driver, long lsf_job_id) {
     char* hostnames = stringlist_alloc_joined_string(hosts, ", ");
     fprintf(stderr, "%s blacklisting nodes %s.\n", __func__, hostnames);
 
+    fprintf(stderr,
+            "%s Realization %ld seems to have failed as a result of LSF node failure.\n",
```
joakim-hove: This error message/warning is for the user and not for developers, no reason to add the leading `__func__`!

markusdregi: Aight

PR #1449, opened 2017-03-21
Hunting compilation warnings

markusdregi:
ERT generates a number of compilation warnings when compiling, many of them due to variables that are never used etc.


markusdregi:
ERT now compiles with the default compilation settings without any warnings on my machine locally. It seems like the Travis setup is even stricter (having more compilation flags) and I will come back in the future and try to tackle these as well.

joakim-hove:
Beautiful

pgdr:
Good job!

PR #1451, opened 2017-03-22
torque status read more robust.

joakim-hove:
OK - the next time I will add tests ....

### DIFF
--- a/libjob_queue/src/torque_driver.c
+++ b/libjob_queue/src/torque_driver.c
@@ -447,67 +447,103 @@ void * torque_driver_submit_job(void * __driver,
   }
 }
 
-static char* torque_driver_get_qstat_status(torque_driver_type * driver, char * jobnr_char) {
-  char * status = util_malloc(sizeof (char)*2);
+/**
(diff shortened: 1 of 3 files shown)


libjob_queue/src/torque_driver.c
```diff
@@ -490,19 +507,22 @@ static char* torque_driver_get_qstat_status(torque_driver_type * driver, char *
...
-    result = JOB_QUEUE_DONE;
-  } else if (strcmp(status, "C") == 0) {
-    result = JOB_QUEUE_DONE;
-  } else if (strcmp(status, "Q") == 0) {
-    result = JOB_QUEUE_PENDING;
-  } else {
-    fprintf(stderr, "%s: Unknown status found (%s), expecting one of R, E, C and Q.\n", __func__, status);
-    result = JOB_QUEUE_STATUS_FAILURE;
+
+  if (status) {
```
pgdr: I would prefer to have early exit here ... `if (!status) { print; result = failure; free; return; } else ...`

But that's not a blocker, so I think we can just merge this.

PR #1454, opened 2017-03-23
Double containment for points on cell faces

markusdregi:
Points that lie on a cells face will be reported as contained in all cells sharing this face. The code for handling corner cases only handles corners..

As a first step I will implement functionality for testing whether a point lies on any of the "sides" of a cell (a cell is considered to have at most 6 sides, each one consisting of either two triangles or one rectangle). One can then use this logic instead for a direct compare to decide whether a point is indeed one of the cells corners (in particular, if a point lies on three of a cells sides it forms one of the eight corners).

### DIFF
--- a/libecl/src/ecl_grid.c
+++ b/libecl/src/ecl_grid.c
@@ -3952,11 +3952,6 @@ bool ecl_grid_cell_contains_xyz3( const ecl_grid_type * ecl_grid , int i, int j 
   return true;
 }
 
-#undef NOT_ON_FACE
-#undef BELONGS_TO_CELL
-#undef BELONGS_TO_OTHER
-
-
 
 bool ecl_grid_cell_contains_xyz1( const ecl_grid_type * ecl_grid , int global_index, double x , double y , double z) {
   int i,j,k;
--- a/libecl/src/ecl_grid.c
+++ b/libecl/src/ecl_grid.c
@@ -500,43 +500,44 @@ which splits the lower face along the 0-3 diagnoal and the upper face
 along the 5-6 diagonal.
 */
 
+
 static const int tetrahedron_permutations[2][12][3] = {{
-                                                        // Top
+                                                        // K-
                                                         {0,1,2},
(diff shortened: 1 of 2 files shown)


libecl/src/ecl_grid.c
```diff
@@ -500,30 +500,46 @@ which splits the lower face along the 0-3 diagnoal and the upper face
 along the 5-6 diagonal.
 */
 
-static const int tetrahedron_permutations[2][12][3] = {{{0,1,2},
+static const int tetrahedron_permutations[2][12][3] = {{
+                                                        // Top
```
joakim-hove: Instead of 'top / bottom / north / south / ...' you should use `I+, I-, J+, J-, K+`and `K-` - those are *reasonably* well defined.  See [this](https://github.com/Ensembles/ert/blob/master/libecl/src/ecl_grid.c#L449)  for the numbering of corners; also need to define how e.g. `{0,1,2}` should be related to a direction - it is not obvious.

Alternatively the strings can just be removed again!

joakim-hove:
Impressive - I did not really understand it; but I guess it should be merged!

libecl/src/ecl_grid.c
```diff
@@ -3777,145 +3894,67 @@ bool ecl_grid_compare(const ecl_grid_type * g1 , const ecl_grid_type * g2 , bool
...
+    if (sign * point3_plane_distance(p0 , p1 , p2 , &p ) < 0)
       return false;
-    }
   }
-}
 
+  return true;
+}
 
+#undef NOT_ON_FACE
```
joakim-hove: The symbols `NOT_ON_FACE, BELONGS_TO_CELL` and `BELONGS_TO_OTHER` are defined as an enum - with file scope, `#undef` here has no consequences - essentially a bug; alternatively you could use 
```C
#define NOT_ON_FACE  1
#define BELONGS_TO_CELL 2
#define BEONGS_TO_OTHER 3
```
where the enum definition is now.

Issue #1370, opened 2016-12-15
Test failure in python.tests.ert.enkf.runpath_list is an impediment


pgdr:
The test `python.tests.ert.enkf.runpath_list`, is randomly failing.  In `enkf_util` we assert that the _file type_ of a given buffer is of type `FIELD (104)` but it is given as `INVALID (0)`:
```c
  if (file_type != target_type) 
    util_abort("%s: wrong target type in file (expected:%d  got:%d)  - aborting \n",
               __func__ , target_type , file_type);
```

This sporadic test failure is beginning to be an impediment, especially when it delays nightly builds, and is generally causing noise on PRs that are fine and completely unrelated to this issue.

The following is the output from `ctest` and the corresponding Ert traceback.

```ctest
248/303 Test #249: python.tests.ert.enkf.runpath_list
                   .............................. ***Exception: Other  3.13 sec
```

```
Abort called from: enkf_util_assert_buffer_type (../libenkf/src/enkf_util.c:106)

Error message: enkf_util_assert_buffer_type: wrong target type in file
(expected:104 got:0) - aborting
```

```trace
#00 ????
#01 util_abort__(..)                                                   in ???
#02 enkf_util_assert_buffer_type(..)                                   in ???
#03 field_read_from_buffer(..)                                         in ???
#04 ????
#05 enkf_state_fread(..)                                               in ???
#06 enkf_state_init_eclipse(..)                                        in ???
#07 enkf_main_icreate_run_path(..)                                     in ???
#08 enkf_main_create_run_path(..)                                      in ???
#09 ffi_call_unix64(..)                                                in ???
#10 ffi_call(..)                                                       in ???
#11 _ctypes_callproc(..)                                               in ???
```

Note that
```c
typedef enum {INVALID          = 0   ,
              IMPL_TYPE_OFFSET = 100 ,
              FIELD            = 104 ,
...
              CONTAINER        = 115 } ert_impl_type;
```

pgdr:
And here's the annoying part, of course; ten minutes later:
```
99% tests passed, 1 tests failed out of 303

Label Time Summary:
Python    = 582.99 sec

Total Test time (real) = 216.49 sec

The following tests FAILED:
	249 - python.tests.ert.enkf.runpath_list (OTHER_FAULT)
Errors while running CTest
Build step 'Execute managed script' marked build as failure
```

jepebe:
It would be nice with a stream like Python class that was synchronized with the C handle...

[...]

PR #1455, opened 2017-03-23
On flaky test runpathlist

pgdr:
Viz  #1370

I added an explicit flaky test.  We need to find out why it occasionally fails.  It is not clear whether this is a problem with python, bindings or in c

The test uses `snake_oil.ert` and somewhere this function is invoked:
```c
void enkf_util_assert_buffer_type(buffer_type * buffer, ert_impl_type target_type) {
  ert_impl_type file_type;
  file_type = buffer_fread_int(buffer);
  if (file_type != target_type) 
    util_abort("%s: wrong target type in file (expected:%d  got:%d)  - aborting \n",
               __func__ , target_type , file_type);
}
```
with `target_type = FIELD (104)`, but it is apparently of type `INVALID (0)`.

### DIFF
--- a/python/tests/ert/enkf/test_runpath_list.py
+++ b/python/tests/ert/enkf/test_runpath_list.py
@@ -13,11 +13,7 @@ class RunpathListTest(ExtendedTestCase):
         # enkf_util_assert_buffer_type: wrong target type in file (expected:104 got:0)
         test_path = self.createTestPath("local/snake_oil_field/snake_oil.ert")
         with ErtTestContext("runpathlist_basic", test_path) as tc:
-            ert = tc.getErt()
-            ens_size = ert.getEnsembleSize()
-            runner = ert.getEnkfSimulationRunner()
-            mask = BoolVector(initial_size=ens_size, default_value=True)
-            runner.createRunPath(mask, 0)
+            pass
 
     def test_runpath_list(self):
         runpath_list = RunpathList('')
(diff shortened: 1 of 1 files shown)


joakim-hove:
This is a good iniative. 

When storing `enkf_node` instances on disk they are stored in a `buffer_type` instance; the storage starts with an integer value from this [enum](https://github.com/Ensembles/ert/blob/master/libenkf/include/ert/enkf/enkf_types.h#L94) and then the  ènkf_node` will call upon the underlying datatype to serialize itself. The failure here is that when deserializeing the type id comes out as zero and not 104 as it should. 

As part of the test context the file in question has just been copied to `/tmp/xxx` - mu hunch is that that something has not been syncronized properly for this copy operation.

python/tests/ert/enkf/test_runpath_list.py
```diff
@@ -8,6 +8,17 @@
 
 class RunpathListTest(ExtendedTestCase):
 
+    def test_an_enkf_runpath(self):
+        # TODO this test is flaky and we need to figure out why.  See #1370
+        # enkf_util_assert_buffer_type: wrong target type in file (expected:104 got:0)
+        test_path = self.createTestPath("local/snake_oil_field/snake_oil.ert")
+        with ErtTestContext("runpathlist_basic", test_path) as tc:
```
joakim-hove: Actually I *trhink* this should be enough to reproduce the flaky behaviour:
```python
test_path = self.createTestPath("local/snake_oil_field/snake_oil.ert")
with ErtTestContext("runpathlist_basic", test_path) as tc:
     pass
```

pgdr: Thanks, I will try that version.

pgdr: You are right.  That was indeed a minimal example.

PR #1456, opened 2017-03-27
Eclipse point containment for concave cells

markusdregi:
The work of this PR is twofold. First, we develop a synthetic grid generator that can create much more evil grids then the current EclGrid.createRectangular. Among others, one can create grids with:

- cells formed as trapezoids,
- cells that are concave,
- cells with overlapping corners,
- faults and
- numerous variations of non-parallel pillars.

We also reimplemented the containment test to handle concave cells and the generator described above is used to test this new code. The point containment test is now done as follows:
 - we decompose the cell into 5 carefully selected tetrahedrons according to the decomposition method
of the faces,
 - after which we test for containment in these 5 tetrahedrons by standard techniques.


joakim-hove:
> The volume of an Eclipse grid cell can be approximated in the following manner

ehhh yes - but why? I don't really understand the purpose of this PR?

markusdregi:
Ah, maybe I was not clear enough. The main purpose is to ensure that the computation of cell volume and point containment is consistent. Hence, the added tests that approximates the volume via the point containment.

The tests fails for non-convex cells and hence this should be fixed. Right now, it also fails for trapezoidal cells, i.e. layers that looks like waves. But this will be ok when the cell_contains_faces branch is merged.

joakim-hove:
>Ah, maybe I was not clear enough. The main purpose is to ensure that the computation of cell volume and point containment is consistent.

OK - but is that even a goal; or realistic - for cells where some of the tetrahedrons contribute with negative volume?

pgdr:
>> ensure that the computation of cell volume and point containment is consistent.

> OK - but is that even a goal;

Obviously yes!

> or realistic - for cells where some of the tetrahedrons contribute with negative volume?

Some _tetrahedrons_ might contribute negatively, but a _cell_ should never have a negative volume, and thus if we compute cell volumes and cell containments based on _cells_ it sounds realistic.

joakim-hove:
OK - this looks good (or frankly - I am not really up to reviewing it). My only wish is that new - quite advanved - functionality for creating grids be split into a separate fully static class - e.g. `GridGenerator`.

markusdregi:
@joakim-hove There is now an EclGridGenerator.

PR #1458, opened 2017-03-28
Add option to add surface to local context

pgdr:
This PR exposes the functionality `add_surface` that takes a `geo_region` to the type `LocalDataset` in python.  The functionality has always been accessible in the underlying `c` libraries (`libgeometry` and `libenkf`), but hasn't been exposed until now.

Adds some missing tests.

Please merge #1465 first.


pgdr:
I broke the PR, will make a new one.

PR #1460, opened 2017-03-28
Use monitor when calculating single vintage distance, not base

kjellkongsvik:
Ganpan said that makes more sense. I Agree.


joakim-hove:
Personally I think the current implememtation is the most logical; currently this will work:

```python
z1 = evalGertsma("S1" , None, .... )
z2 = evalGertsma("S2" , None, ... )
dz = evalGertsma( "S1" , "S2", ...)

dz = z1 - z2
```
That will get an overall sign-change with this alternative.  In addition there are people using the subsidence code as-is, no reason to surprise them?

So - unless there is a very good reason to merge this - I would prefer to close it.

kjellkongsvik:
I see it will be inconsistent with subsidence yes. Ok. Close and we will manage

PR #1469, opened 2017-03-30
Geo region and geo pointset

pgdr:


libgeometry/src/geo_surface.c
```diff
@@ -116,28 +91,20 @@ static void geo_surface_init_regular( geo_surface_type * surface , const double
...
 
     if (index == surface->nx * surface->ny) {
       double extra_value;
       int fscanf_return = fscanf( stream , "%lg" , &extra_value);
-
-      /* Check that there is not more data dangling at the end of the file. */
-      if (fscanf_return == EOF)
-        OK = true;
-      break;
+      return fscanf_return == EOF; // no more data dangling at the end of the file.
```
joakim-hove: Can I wish for a paranthesis: `return (fsacnf_return == EOF);`?

python/python/ert/geo/cpolyline.py
```diff
@@ -41,7 +41,7 @@ class CPolyline(BaseCClass):
     _equal              = GeoPrototype("bool     geo_polygon_equal( geo_polygon , geo_polygon )")
 
 
-    def __init__(self, name = None , init_points = []):
```
joakim-hove: I guess that the `init_points` variable should be a list - i.e. `[]`; not that it matters much for the default though.

python/python/ert/geo/surface.py
```diff
@@ -234,12 +239,51 @@ def getXY(self, index):
...
+                i_idx += self.getNX()
+            if j_idx < 0:
+                j_idx += self.getNY()
+            if 0 <= i_idx < self.getNX() and 0 <= j_idx < self.getNY():
+                idx = j_idx * self.getNX() + i_idx
+            else:
+                fmt = 'Index error: i=%d not in [0,nx=%d) or j=%d not in [0,ny=%d).'
+                raise IndexError(fmt % (i, nx, j, ny))
+
+        from sys import stdout
```
joakim-hove: I guess this code is left-over from debugging?

python/tests/core/geometry/test_geo_pointset.py
```diff
@@ -0,0 +1,28 @@
+from ert.geo import GeoPointset, Surface
+from ert.test import ExtendedTestCase, TestAreaContext
+
+
+class GeoPointsetTest(ExtendedTestCase):
+
+    def test_init(self):
+        gp = GeoPointset()
+        print(repr(gp))
```
joakim-hove: Left-over `print`?

python/tests/core/geometry/test_geo_region.py
```diff
@@ -0,0 +1,16 @@
+from ert.geo import GeoRegion, GeoPointset
+from ert.test import ExtendedTestCase, TestAreaContext
+
+
+class GeoRegionTest(ExtendedTestCase):
+
+    def test_init(self):
+        ps = GeoPointset()
+        gp = GeoRegion(ps)
+        print(repr(gp))
```
joakim-hove: `print`

pgdr:
will cherry-pick into another branch

joakim-hove:
> will cherry-pick into another branch

You should get away with an interactive rebase on `upstream/master` - that was what we did here: https://github.com/Ensembles/ert/pull/1470

There is even the "git rebase onto" which should be able to pull this off in one go.

joakim-hove:
http://stackoverflow.com/questions/29914052/i-cant-understand-the-behaviour-of-git-rebase-onto

PR #1475, opened 2017-04-03
Fixed bug with RPC server and GEN_KW values.

joakim-hove:
????

Testing as we speak.


joakim-hove:
OK - this has been user-tested; merging *and* backporting to 2.1

PR #1477, opened 2017-04-03
Select polygon and halfspaces

pgdr:
Functionality to
1. select polygons from surface
2. select halfspace from surface


joakim-hove:
I'll close this - and continue review on #1478.

PR #1478, opened 2017-04-04
Documentation: Refactored `user/localization` RST.

pgdr:
Refactored:
* removed absurdly long lines
* removed trailing whitespace
* removed abused `|` character; it's purpose is to inhibit reflow of paragraph
* added
  ```
   ``double backticks``
  ```
  around code related text

This is based on #1477.

### DIFF
--- a/python/python/ert/geo/geo_region.py
+++ b/python/python/ert/geo/geo_region.py
@@ -45,31 +45,21 @@ class GeoRegion(BaseCClass):
             super(GeoRegion, self).__init__(c_ptr)
         else:
             raise ValueError('Could not construct GeoRegion from pointset %s.' % pointset)
-        self.select_polygon_fn = {
-            (True, True): self._select_inside_polygon,
-            (False, True): self._select_outside_polygon,
-            (True, False): self._deselect_inside_polygon,
-            (False, False): self._deselect_outside_polygon
-        }
-        self.select_halfspace_fn = {
-            (True, True): self._select_above_line,
-            (False, True): self._select_below_line,
-            (True, False): self._deselect_above_line,
-            (False, False): self._deselect_below_line
-        }
 
 
     def getActiveList(self):
         return self._get_index_list()
 
-    def select_polygon(self, polygon, inside=True, select=True):
+    def _assert_polygon(self, polygon):
         if not isinstance(polygon, CPolyline):
             raise ValueError('Need to select with a CPolyline, not %s.'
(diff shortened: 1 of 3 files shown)


joakim-hove:
Very good with the `rst` documentation.

python/python/ert/geo/geo_region.py
```diff
@@ -41,10 +45,68 @@ def __init__(self, pointset, preselect=False):
...
+            x2, y2 = map(float, p2)
+        except Exception as err:
+            err_msg = 'Select with pair ((x1,y1), (x2,y2)), not %s (%s).'
+            raise ValueError(err_msg % (line, err))
+        x1x2_ptr = cpair(x1, x2)
+        y1y2_ptr = cpair(y1, y2)
+        selector_fn = self.select_halfspace_fn[(above, select)]
+        selector_fn(x1x2_ptr, y1y2_ptr)  # note that it takes xx, yy
+
+    def select_inside(self, polygon):
```
joakim-hove: At least for the polygon the  gain from having the `(bool, bool)` tuple based function table is quite limited; these functions could just call the select function directly:

```
def select_inside(self , polygon):
       self._select_inside_polygon( polygon )
```
possibly with a @assert_polygon decorator[1]? Or  just leave it as it is if you think that is best.

[1]: Yes - I think decorators is a quite cool concept which we do not use very much.

python/python/ert/geo/geo_region.py
```diff
@@ -41,10 +45,68 @@ def __init__(self, pointset, preselect=False):
...
+            (False, True): self._select_below_line,
+            (True, False): self._deselect_above_line,
+            (False, False): self._deselect_below_line
+        }
+
 
     def getActiveList(self):
         return self._get_index_list()
 
+    def select_polygon(self, polygon, inside=True, select=True):
```
joakim-hove: Ahhh  - these f...ng names. Initially the ert codebase used `snake_case` - then someone came along from the Java part of the woods and said that `cameCase` was all the rage, and we started a gradual `snake_case` &rightarrow; `camelCase` transition. Since then it seems `camelCase` as fallen out of favor - and we are quite stable at 50% of each.

Personally I have no strong opinion, but:

**I really think we should strive for consistency. The current approach with gradually moving the code in either direction just does not work. We need to agree on or the other and make an all out effort to ensure consistency.**

pgdr:
addressed comments

PR #1480, opened 2017-04-04
__len__(EclGrid) -> total number of cells.

joakim-hove:
Does this make sense?


joakim-hove:
> but then I thought we should also have a __getitem__, but wasn't sure whether it would be safe to assume people would use grid[key] for 0 <= key < nx*ny*nz in a safe and understandable manner.

I guess `grid[cell_index]` could return a cell object - makes some sense. Not now though.

PR #1481, opened 2017-04-05
Blacklist fixup

joakim-hove:


joakim-hove:
I will squash for simple cherry-picking before merge.

pgdr:
Just wanted to mention that Matthew designed the output message, so take that into account, before you change it.

joakim-hove:
> Just wanted to mention that Matthew designed the output message, so take that into account, before you change it.

OK thank you - I'll ask him to take a look.

PR #1482, opened 2017-04-06
OTS: base-monitor in subsidence

kjellkongsvik:


kjellkongsvik:
Had a detailed review with Ganpan. As this is the subsidence it should be base-monitor. External timeshift will change sign explicitly

PR #1483, opened 2017-04-06
Nnc geometry

joakim-hove:


joakim-hove:
Moved to: https://github.com/Statoil/libecl/pull/1

PR #1485, opened 2017-04-06
Disable python for static

joakim-hove:
If the `-DBUILD_SHARED_LIBS=OFF` option is passed the Python wrappers are automatically disabled.


CMakeLists.txt
```diff
@@ -139,7 +139,8 @@ else()
       message(FATAL_ERROR "The full ERT application must be built with shared libraries")
    endif()
    if (BUILD_PYTHON)
-      message(FATAL_ERROR "The Python wrappers require shared libraries")
+      message("The Python wrappers require shared libraries - Python will be disabled")
```
pgdr: Could issue a `WARNING` but not a pressing issue

PR #1499, opened 2017-06-19
Pandas drops duplicates

markusdregi:
Backporting the following PR: https://github.com/Statoil/ert/pull/34


joakim-hove:
Will you also add a commit raising the micro version to 2.1.8.

